PR #22, opened 2015-04-08
Add init that accepts Optional and a failWith Error.

rnapier:
Thoughts? I've been thinking of several ways to implement this, but this approach seems nicest so far.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -14,6 +14,10 @@ public struct Result<T, Error>: EitherType, Printable, DebugPrintable {
 		either = .left(error)
 	}
 
+    /// Constructs a result from an Optional, failing with `Error` if `nil`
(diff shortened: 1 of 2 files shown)


robrix:
Nice—I dig the approach. I wonder if `failWith:` is the clearest label? Perhaps `otherwise:`?

rnapier:
`otherwise` feels like `??` which makes me think that it would take `T` rather than `Error`.

`orError:`? Though `or...` feels a bit very-old-school ObjC :D (like `getObjects:andKeys:`).

    Result(doSomething(),  
        orError: mkError("Why you fail?"))`

Just `error:` didn't feel right, because it's one or the other; not both.

So maybe `otherwise:` works. I don't think I have a strong opinion here.

robrix:
Hm, good points. `failWith:` is fine :+1:

PR #24, opened 2015-04-10
Remove .xcworkspace

felixjendrusch:
Using just a `.xcodeproj` works around [`xcodebuild` nondeterministically selecting implicit dependencies for the wrong platform in `.xcworkspace`s](http://www.openradar.me/20490378).

Depends on https://github.com/robrix/Either/pull/27. See also https://github.com/Carthage/Carthage/issues/418.


robrix:
Per [discussion](https://github.com/Carthage/Carthage/issues/418#issuecomment-91495670), this won’t work for duplicated transitive dependencies. I can understand that you can’t wait for the Xcode team, so in the meantime I’d suggest either working off your branch or renaming the products per-platform (i.e. `Result-iOS.framework`).

PR #26, opened 2015-04-16
Setting APPLICATION_EXTENSION_API_ONLY to YES

NachoSoto:
Per the documentation:

```
When enabled, this causes the compiler and linker to
disallow use of APIs that are not available
to app extensions and to disallow linking to frameworks
that have not been built with this setting enabled.
```

Since we're not using any of these APIs, we can safely turn this setting on.

Related: https://github.com/ReactiveCocoa/ReactiveCocoa/pull/1891
This is in preparation for https://github.com/ReactiveCocoa/ReactiveCocoa/issues/1887.

Note: I set this at the project level so that it affects both iOS and Mac.


robrix:
:+1:

PR #29, opened 2015-04-19
#nodependencies

### LINKED ISSUE #28: Proposal: remove Either dependency
It's indisputable that `Result` is a kind of `Either`, but every added dependency is an additional burden for consumers, who I'm not convinced will use the `EitherType` functionality directly.

Furthermore, the Either [README](https://github.com/robrix/Either/blob/8d3f69d109436f74f68e4828a75be87eea12fa4b/README.md#use) even suggests that its functionality can be trivially added as an extension upon a pre-existing `Result` type. To me, this means that consumers should not be _required_ to build and link Either just to use a standardized `Result` type.

c.f. [the conversation from #reactivecocoa](https://reactivex.slack.com/archives/reactivecocoa/p1429157460000503)

/cc @nachosoto @saniul @tony

robrix:
Fixes #28.


robrix:
Ready for review.

rnapier:
:+1:

robrix:
@jspahrsummers Any concerns with this approach?

robrix:
It occurs to me we could make `Box` private:

- make a private `ResultState` `enum` with `analysis` and nothing else
- make `Result` a public `struct` with the real API

jspahrsummers:
:100: All of the above sounds awesome. Thanks @robrix!

Hiding the `Result` cases might make pattern matching slightly more difficult, but `Box` kind of impaired that anyways (since you can't match through a box, IIRC). And maybe we could implement a custom `~=`? Not sure how that works.

robrix:
Correct: you can’t match through anything but an `enum` or tuple. To the best of my knowledge, `~=` is limited to pass/fail, so that wouldn’t allow us to destructure, but we could maybe match against expected error/success values; and `analysis` recovers destructuring.

Gonna hide `Box` away now.

robrix:
Alright, it’s done :mahjong:

mdiep:
Can we get a release with this change? :grin:

robrix:
@mdiep :+1:

robrix:
@mdiep: https://github.com/antitypical/Result/releases/tag/0.3

mdiep:
Thank you! :pray:

robrix:
:bow: you’re welcome!

PR #31, opened 2015-04-21
Add State.swift to iOS target

jcavar:
Hi, I am getting errors when trying to build iOS target. State.swift was not assigned to iOS target so I have just ticked box.


robrix:
Oof. Good fix.

Issue #41, opened 2015-05-12
Add `recover` and `recoverWith`


Thomvis:
Would you be open to adding `recover` and `recoverWith` to `Result`. They're similar to `map` and `flatMap` except that they are given closures that run when the `Result` failed and then provide an opportunity to make it succeed. The implementation of `recover` would actually be identical to the `??` operation.

If you are favorable of this change, I'd like to make a PR.

For more context, see https://github.com/Thomvis/BrightFutures/pull/49, where we're discussing to replace a custom Result implementation with this library and identify the functions discussed here as the biggest missing piece.

The naming comes from Scala (http://www.scala-lang.org/api/2.10.1/index.html#scala.concurrent.Future), at least that's where I got it from.

robrix:
Thanks for filing this!

I’m of two minds. Against:

- I really like reinforcing use of the extant `??` operator—overloading is very much in Swift’s idiom, and the analogy between `Result`’s `??` and `Optional`’s is clear, and supportive of general use of the language. Adding `recover`/`recoverWith` dilutes that somewhat, and so I’m a bit reluctant.

For:

- `??` doesn’t cover `recoverWith`, and an `??` overload taking a function on the right hand side feels like it’s too distinct in usage from the extant overloads to be reasonably named `??`, and I don’t much want to define new custom operators.
- `recover` and `recoverWith` are pretty decent names with some precedent, and names are searchable where operators are not; for which reason (in part) we have `flatMap` as well as `>>-`.

On the balance, I think it’d make a fine addition. My only request is that the documentation for `recover` mention the equivalence with `??`.

Thank you kindly!

robrix:
Fixed by #42. I’ll tag in a moment.

Thomvis:
@robrix Thanks for the fast response & thanks @nvh for the PR :star2:

nvh:
:tada: :heart:

robrix:
You’re very welcome :smile:

PR #42, opened 2015-05-13
Add `recover` and `recoverWith`

nvh:
This implements the changes discussed in #41

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -73,6 +73,18 @@ public enum Result<T, Error>: Printable, DebugPrintable {
 			ifSuccess: transform,
 			ifFailure: Result<U, Error>.failure)
 	}
+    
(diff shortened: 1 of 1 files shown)


Result/Result.swift
```diff
@@ -73,6 +73,18 @@ public enum Result<T, Error>: Printable, DebugPrintable {
...
 			ifFailure: Result<U, Error>.failure)
 	}
+    
+    /// Returns `self.value` if this result is a .Success, or the given value otherwise. Equivalent with `??`
+    public func recover(value: T) -> T {
+        return self.value ?? value
+    }
+    
+    /// Returns this result if it is a .Success, or the given result otherwise. Equivalent with `??`
+    public func recoverWith(result: Result<T,Error>) -> Result<T,Error> {
```
robrix: AIUI this one is not equivalent with `??`, because `??` doesn’t take a function on the right-hand side.

robrix:
:+1:

PR #46, opened 2015-05-14
Lazily evaluated result coalescing

robrix:
`??` used `@autoclosure` but since `recover`/`recoverWith` did not, it made no difference: the right hand operand would always be evaluated eagerly.

The fix is to use `@autoclosure` for `recover` and `recoverWith`, which I believe is correct for their semantics as well.

h/t @esttorhe: https://twitter.com/esttorhe/status/598962925685080064

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -73,8 +73,8 @@ public enum Result<T, Error>: Printable, DebugPrintable {
 	}
 	
 	/// Returns `self.value` if this result is a .Success, or the given value otherwise. Equivalent with `??`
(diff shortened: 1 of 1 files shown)


esttorhe:
:clap: great response time.

With this in place I think I can rewrite my code better now :bow:

Thanks :tada:

robrix:
@esttorhe: released: https://github.com/antitypical/Result/releases/tag/0.4.3

esttorhe:
:+1:  let me hit `carthage update` then :)

PR #47, opened 2015-05-18
Implement conjunction operator

ikesyo:
This can be used here for instance: https://github.com/Carthage/ReactiveTask/blob/94d0725de7d116019a704642c6566cd02698bd5f/ReactiveTask/Task.swift#L266-L267

```swift
SignalProducer(result: Pipe.create() &&& Pipe.create())
```

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -141,6 +141,10 @@ public func != <T: Equatable, Error: Equatable> (left: Result<T, Error>, right: 
 	return !(left == right)
 }
 
+/// Returns a Result with a tuple of `left` and `right` values if both are `Success`es, or re-wrapping the error of the earlier `Failure`.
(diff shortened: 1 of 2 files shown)


robrix:
I’ve named the analogous operator in Prelude `&&&`. What do you think about that name instead?

robrix:
(To be clear, I’m asking more than suggesting :smile:)

robrix:
(And thank you for the PR! :bow:)

ikesyo:
I'm OK with that change. 👌 If do so, would I add the `&&&`  operator declaration here directly, not depend on Prelude, OK?

robrix:
Absolutely, let’s not add a dependency for this. Let’s just make sure that it has the same precedence and associativity as Prelude uses.

ikesyo:
Thanks for the clarification. :pray: I've done it.

robrix:
:sparkles: Looks great!

ikesyo:
Thanks! :sparkling_heart:

PR #48, opened 2015-05-21
Add an optional message parameter for NSLocalizedDescriptionKey to Result.static.error()

ikesyo:


robrix:
My intention with the errors generated by this API was that they would be the underlying errors of higher-level errors generated by the consumer. That kind of obviates the need for a localized description in this API.

That said, wrapping errors isn’t super convenient.

@antitypical/result: Anybody else have any thoughts?

jspahrsummers:
I'm hoping to avoid `NSError` as much as possible in Swift APIs, so anything that reduces the need to create them myself sounds good.

robrix:
That’s good enough for me :+1:

robrix:
Thanks for the PR, @ikesyo!

ikesyo:
Thank you for merging! :metal:

PR #53, opened 2015-05-28
Add Error transformations.

swizzlr:
This commit adds new overloads of `map` and `flatMap` that operate on the error
side of the Result.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -65,13 +65,25 @@ public enum Result<T, Error>: Printable, DebugPrintable {
 		return flatMap { .success(transform($0)) }
 	}
 
+	/// Returns a new Result by mapping `Failure`s’ values using `transform`, or re-wrapping `Success`es’ values.
+	public func map<NewError>(@noescape transform: (Error) -> NewError) -> Result<T, NewError> {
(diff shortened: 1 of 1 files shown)


swizzlr:
I'm using this as an extension right now but it seemed to me you may be interested in it. I'm not precious about getting it merged if you fear it bloats the API.

robrix:
I’m not hugely averse to adding these, but I think it might be worth avoiding the overloads of `map`/`flatMap`.

I think the `flatMap` added here is akin to RAC’s `catch`?

/cc @antitypical/result Any thoughts?

mdiep:
Agree that overloading doesn't seem ideal. But I'm not sure if there's any precedent.

swizzlr:
RAC has `mapError` and no `flatMap` that operates on errors. Perhaps `mapError` alone would make more sense?

jspahrsummers:
@swizzlr `catch` is `flatMap` over errors.

swizzlr:
oooooh. I'm not even using this anymore, since I'm using RAC heavily, and since it's relatively trivial to implement, I'll close this unless anyone thinks this API is helpful.

PR #54, opened 2015-06-05
[RFC] Use Runes as an optional dependency for functional operators

gfontenot:
Been thinking about this a lot lately, and wanted to open up a discussion about it.

Issues like #49 seem like a smell to me. Needing to maintain duplicate and parallel operator definitions in order to maintain compatibility with other libraries screams "shared dependency" to me. Ideally, I think [Runes](https://github.com/thoughtbot/runes) (or a similar lib) would define the operators, and then `Result` would define implementations for the types it exposes. That would allow the operators to have a single point of definition, and guarantee consistency for things like precedence.

Obviously, it's not that simple. Adding Runes as a dependency would mean needing to expose another framework that users would need to link against, even if they never use it. I think we're all in agreement that a lib like this should have as few dependencies as possible. `Box` makes sense, but adding in a dependency on `Runes` (especially after we've already pulled out the dependencies on `Either` and `Prelude`) doesn't.

Ideally, we'd be able to do some sort of optional dependency. It would be awesome to be able to say "Is Runes available? If so, here's this implementation of `flatMap` for result that you can use". (It would be _more_ awesome to be able to define some sort of generic type for this. Like a higher kind of type. idk) But unfortunately Apple hasn't provided a way to do that.

As a temporary solution, what if the duplicate operators were moved into a second framework that included Runes as a dependency, leaving the basic `Result` framework pure?

__PROS__
- Users that never use `>>-` or any of the other operators would never see them, and would never need to worry about them
- Users that _do_ use `>>-` and so forth wouldn't have to worry about conflicting operator definitions.
- Using a common dependency that implements these same operators for `Optional` and `Array` would theoretically increase exposure to these operators, driving education and adoption.

__CONS__
- This would require users that want `>>-` to import a module that is named something other than `Result`. `ResultRunes` or `ResultOperators` or whatever. That's fairly gross.
- When using Carthage to install `Result`, users would be left with 5(!) frameworks in their build directory. They would then have to know which ones to link to get what they want.

As a side note, this would be able to be implemented for CocoaPods by using subspecs. CP users would then only need to import `Result` no matter what, using a spec similar to this:

```ruby
Pod::Spec.new do |s|
  s.name         = 'Result'
  s.version      = '0.4.3'
  s.summary      = 'Swift type modelling the success/failure of arbitrary operations'

  s.homepage     = 'https://github.com/antitypical/Result'
  s.license      = 'MIT'
  s.author       = { 'Rob Rix' => 'rob.rix@github.com' }
  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version.to_s }
  s.source_files  = 'Result/Result.swift'
  s.requires_arc = true
  s.ios.deployment_target = '8.0'
  s.osx.deployment_target = '10.9'

  subspec 'Runes' do |subspec|
    subspec.source_files = 'Result/Runes.swift'
    subspec.dependency 'Runes', '>= 1.2.2'
  end
end
```

CP users would then be able to declare `pod 'Result/Runes'` and they would get both.

I don't think this PR should be used as the final implementation by the way, but meant it more as an example of how this _could_ work in the service of conversation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "Carthage/Checkouts/Runes"]
+	path = Carthage/Checkouts/Runes
+	url = https://github.com/thoughtbot/Runes.git
--- a/Cartfile
+++ b/Cartfile
@@ -1,0 +1,1 @@
+github "thoughtbot/Runes" "gf-framework-dependency"
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,0 +1,1 @@
+github "thoughtbot/Runes" "269bee6e51b4b9de5c2658f1927a8fa6a1a731f8"
--- a/Result.xcodeproj/project.pbxproj
+++ b/Result.xcodeproj/project.pbxproj
@@ -15,6 +15,8 @@
 		D45480981A957465009D7229 /* Result.swift in Sources */ = {isa = PBXBuildFile; fileRef = D45480961A957465009D7229 /* Result.swift */; };
 		D45480991A9574B8009D7229 /* ResultTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D454806E1A9572F5009D7229 /* ResultTests.swift */; };
 		D454809A1A9574BB009D7229 /* Result.h in Headers */ = {isa = PBXBuildFile; fileRef = D454805C1A9572F5009D7229 /* Result.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		F80EABB61B28B4640013350E /* Runes.swift in Sources */ = {isa = PBXBuildFile; fileRef = F80EABB51B28B4640013350E /* Runes.swift */; };
+		F80EABB71B28B4640013350E /* Runes.swift in Sources */ = {isa = PBXBuildFile; fileRef = F80EABB51B28B4640013350E /* Runes.swift */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -44,6 +46,7 @@
 		D454807D1A957361009D7229 /* Result.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = Result.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		D45480871A957362009D7229 /* Result-iOSTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = "Result-iOSTests.xctest"; sourceTree = BUILT_PRODUCTS_DIR; };
 		D45480961A957465009D7229 /* Result.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Result.swift; sourceTree = "<group>"; };
+		F80EABB51B28B4640013350E /* Runes.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; name = Runes.swift; path = Carthage/Checkouts/Runes/Source/Runes.swift; sourceTree = SOURCE_ROOT; };
 /* End PBXFileReference section */
 
 /* Begin PBXFrameworksBuildPhase section */
@@ -106,6 +109,7 @@
 			children = (
 				D454805C1A9572F5009D7229 /* Result.h */,
 				D45480961A957465009D7229 /* Result.swift */,
+				F80EABB51B28B4640013350E /* Runes.swift */,
 				D454805A1A9572F5009D7229 /* Supporting Files */,
 			);
 			path = Result;
@@ -311,6 +315,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				D45480971A957465009D7229 /* Result.swift in Sources */,
+				F80EABB61B28B4640013350E /* Runes.swift in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -327,6 +332,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				D45480981A957465009D7229 /* Result.swift in Sources */,
+				F80EABB71B28B4640013350E /* Runes.swift in Sources */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -186,14 +186,6 @@ public func `try`(function: String = __FUNCTION__, file: String = __FILE__, line
 
 // MARK: - Operators
 
-infix operator >>- {
-	// Left-associativity so that chaining works like you’d expect, and for consistency with Haskell, Runes, swiftz, etc.
-	associativity left
-
-	// Higher precedence than function application, but lower than function composition.
-	precedence 100
-}
-
 infix operator &&& {
 	/// Same associativity as &&.
 	associativity left


robrix:
> Obviously, it's not that simple. Adding Runes as a dependency would mean needing to expose another framework that users would need to link against, even if they never use it.

I think using the operators would require importing the framework defining them. Not certain about that tho.

gfontenot:
Oh, right. Using the operators would mean importing _both_ frameworks

I don't think that's a huge deal, to be honest, but maybe I'm the outlier?

robrix:
Forgot to say this earlier, so: Thanks for kicking off this discussion, @gfontenot! :bow:

> I don't think that's a huge deal, to be honest, but maybe I'm the outlier?

I was surprised at how much pushback there was at Result depending on Prelude & Either. Dropping it to zero dependencies (even tho it was almost immediately partially reverted back to one) was frankly disappointing, in that the community as a whole is simply not as accepting of dependencies as you and I are.

So, I worry that we’d be pushing away users who care about this (however unreasonable their concern might seem), which would undermine Result’s goal of ubiquity. (And potentially Runes’ as well if it also used a shared operator dependency?)

At the same time, manually syncing up operator definitions between umpteen frameworks is, frankly, a nuisance at best; at worst it makes for ideal conditions for inconsistencies & ambiguities between dependencies, which is precisely what we’re trying to avoid by writing small ones in the first place.

An optional dependency seems like a pretty reasonable middle ground. Maybe the expense of (potentially) traumatizing whatever set of users lies at the intersection of those who tally dependencies & those who use `>>-` is acceptable.

@antitypical/result: I’d love to hear your thoughts on this.

jspahrsummers:
What about an `Operators.swift` file (not framework) that can be shared and imported via subtrees or submodules, and which declares operator precedence but does not provide operator implementations? Would Swift let us do that?

neilpa:
> Would Swift let us do that?

I believe so, but how would we version this?

jspahrsummers:
Carthage could still be responsible for picking the version—it just shouldn't be something that client applications need to link (i.e., a framework).

robrix:
Even if we _didn’t_ version it automatically, it would make some sense to extract the common operators like that. Kind of a lo-fi solution, which I dig.

@gfontenot What do you think about that? I don’t know if it really solves all of the problems you had in mind here, but it might be a start?

gfontenot:
I _love_ the idea of a micro non-framework powering micro frameworks. Couple of concerns/questions:

- Assuming this file is linked into Runes (defining implementations for Optional/Array) and Result, if they are different versions, do we run into the same compatibility errors? My gut says yes. If that's the case, have we gained anything (other than the obvious Single Point Of Truth)?
- I do think it's important to have a solution that doesn't force maintainers to drop support for a dependency management solution, and I can't think of a way to dynamically include that file with CocoaPods.

neilpa:
> I can't think of a way to dynamically include that file with CocoaPods.

You could still package it as a framework which CocoaPods can use. For those of us using Carthage, it'll act like a header file but instead of `#include` it'll get added to the compiled sources.

jspahrsummers:
> Assuming this file is linked into Runes (defining implementations for Optional/Array) and Result, if they are different versions, do we run into the same compatibility errors? My gut says yes. If that's the case, have we gained anything (other than the obvious Single Point Of Truth)?

IMO the gain is in the _easy maintenance_ of a shared set of operators.

In the framework approach, you'd still have compatibility issues if Result and another library using Runes disagree about what version of Runes to use.

kareman:
I really like the idea of a single shared file for things that various frameworks need to agree upon if they are going to be used together. 

If we would want to have several of these files in a GitHub repository, would it be possible to specify the specific file(s) I am interested in in the Cartfile, as opposed to just the entire repository?

kzaher:
Hi @robrix,

Sorry for dropping in from out of the blue but here are my two cents on this.

In some of my own projects I've made my own result class, and defined operators on it (it was about the same you released Result project). 

The problem is that those operators even though I thought they are were sexy they didn't actually add any value to my  project. They were actually confusing people because people don't know what those symbols mean. Using functions and methods is a lot clearer way IMHO.

In your case writing >>- instead of flatMap saves you 4 characters but also confuses people when they read that code. That actually isn't a significant saving because intellisense and autocomplete are omnipresent.

On top of that there are also problems with naming collisions with other frameworks. 

Yes, you could probably work out some complicated solutions with dependencies, but the original problem of confusing new users and readability will still remain.

One of the comments I got to my project on Twitter was that I'm doing ASCII art with my code, and the guy was right. I've actually removed those operators and the code actually looks much nicer and more readable now with maps and flatMaps.

Anyway, my advice is that you keep Result simple, and just don't define custom operators.

robrix:
> Anyway, my advice is that you keep Result simple, and just don't define custom operators.

Well, kind of too late for that, they’re public & supported API.

jspahrsummers:
> If we would want to have several of these files in a GitHub repository, would it be possible to specify the specific file(s) I am interested in in the Cartfile, as opposed to just the entire repository?

If we were to have several files that make sense to use separately, they should just be in separate repositories.

gfontenot:
> You could still package it as a framework which CocoaPods can use.

I'm not sure that's right, though. If CocoaPods is expecting a framework, that means the file defining the implementations needs to `import Operators` in order to build properly. That's obviously different than just linking the file directly, which is what Carthage/Submodules would do. Unless I'm missing something here?

> IMO the gain is in the easy maintenance of a shared set of operators.
> 
> In the framework approach, you'd still have compatibility issues if Result and another library using Runes disagree about what version of Runes to use.

Yep, good point. I'm cool with this (and am happy to extract the operator definitions out of Runes) assuming these distribution concerns can be addressed.

ikesyo:
> I'm not sure that's right, though. If CocoaPods is expecting a framework, that means the file defining the implementations needs to `import Operators` in order to build properly. That's obviously different than just linking the file directly, which is what Carthage/Submodules would do. Unless I'm missing something here?

Podspecs can handle submodules, so including required source files to a spec and its project is OK (e.g. [Swiftz.podspec](https://github.com/typelift/Swiftz/blob/334e789dc4ad57bd49c37f6796df1c16739fe28f/Swiftz.podspec#L12-L13)). Is it not the case?

gfontenot:
> Podspecs can handle submodules

I literally had no idea. If that's the case, then yeah, we should be able to create Podspecs that work.

gfontenot:
Updated this PR with a better implementation that only adds in `Runes.swift` from the Runes project. I have a [PR opened][runes-36] with Runes to rename Operators.swift -> Runes.swift, since I think this will be better for minimizing conflicts. That PR will also introduce more documentation about how to add Runes as a low-overhead dependency for centralizing operator definitions.

One question: should we use `Cartfile.private` for this? Assuming we're using submodules, Carthage _should_ still check that dependency out, right? That would mean that the user won't ever see Runes.framework at all, which might be better to avoid confusion when linking.

[runes-36]: https://github.com/thoughtbot/Runes/pull/36

gfontenot:
Also, fwiw, the podspec would look like so:

```ruby
Pod::Spec.new do |s|
  s.name         = 'Result'
  s.version      = '0.4.3'
  s.summary      = 'Swift type modelling the success/failure of arbitrary operations'

  s.homepage     = 'https://github.com/antitypical/Result'
  s.license      = 'MIT'
  s.author       = { 'Rob Rix' => 'rob.rix@github.com' }
  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version.to_s, :submodules => true}
  s.source_files  = 'Result/Result.swift', 'Carthage/Checkouts/Runes/Source/Runes.swift'
  s.requires_arc = true
  s.ios.deployment_target = '8.0'
  s.osx.deployment_target = '10.9'
end
```

PR #56, opened 2015-06-09
Add `transform` for functions using `throws`

gfontenot:
This introduces a new higher order function that takes a function that
uses the new `throws` functionality introduced in Swift 2.0 and wraps
it to turn it into a function that returns a `Result` type

hat tip @radex
https://twitter.com/radexp/status/608208346777325568

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -84,8 +84,8 @@ public enum Result<T, Error>: CustomStringConvertible, CustomDebugStringConverti
 			ifFailure: { _ in result() })
 	}
 
-	/// Returns a transformed function that used `throws` to one that uses `Result`
-	public static func transform<T, U>(f: T throws -> U) -> T -> Result<U, ErrorType> {
+	/// Transform a function from one that uses `throw` to one that returns a `Result`
+	public static func materialize<T, U>(f: T throws -> U) -> T -> Result<U, ErrorType> {
 		return { x in
 			do {
 				return .success(try f(x))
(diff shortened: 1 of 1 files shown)


Result/Result.swift
```diff
@@ -84,6 +84,17 @@ public enum Result<T, Error>: CustomStringConvertible, CustomDebugStringConverti
 			ifFailure: { _ in result() })
 	}
 
+	/// Returns a transformed function that used `throws` to one that uses `Result`
```
gfontenot: this is pretty poorly worded now that I'm reading it.

radex:
It would also be nice to add a method that extracts result's value and throws if there's an error.

So essentially: `func extract() throws -> T`

neilpa:
Throwing out `materialize` and `dematerialize` as names for these.

neilpa:
Throwing out `materialize` and `dematerialize` as names for these.

gfontenot:
@radex Solid idea. We'd have to construct a custom `ErrorType` for this though, right?

neilpa:
Now that there's a stdlib `ErrorType` the error param should probably conform to that.

gfontenot:
Agreed. Should that be a separate PR?

neilpa:
Doesn't matter much to me. I thought about it when doing the initial swift 2 PR but forgot until now.

tonyd256:
As a thought, what if this was more like a protocol `Transformable` (name pending) that you could implement. I can see this function being useful to return `Optional`, `Result`, (insert custom type). I know you want to be careful about more dependencies, but a micro framework for this might be interesting.

gfontenot:
Opened separate PRs for `extract` and `ErrorType` conformance.

tonyd256:
Thinking about this more, I don't think my idea makes sense here.

gfontenot:
renamed `tranform` -> `materialize`

PR #57, opened 2015-06-09
Conform error cases to ErrorType

gfontenot:
This is provided by the stdlib, it makes sense for us to conform to it.

As per discussion in #56


robrix:
:+1:

PR #58, opened 2015-06-09
Add `extract` function for deconstruction

gfontenot:
This function extracts the value or throws the error.

As per discussion in #56 /cc @radix
Depends on #57

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -45,6 +45,15 @@ public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStr
 		return analysis(ifSuccess: { _ in nil }, ifFailure: { $0 })
 	}
 
+	public func extract() throws -> T {
+		switch self {
+		case let .Success(value):
+			return value
+		case let .Failure(error):
+			throw error
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


robrix:
#57 has been merged :white_check_mark:

robrix:
Just wanted to reflect upon @neilpa’s [suggestion](https://github.com/antitypical/Result/pull/56#issuecomment-110517858) a bit more:

> Throwing out `materialize` and `dematerialize` as names for these.

I kinda like that due to the harmony with RAC. @antitypical/result, anybody have feels?

gfontenot:
Is there any precedent in other languages that we can pull from? I kind of like `extract` for this, since that sounds like a simple but possibly dangerous operation. `dematerialize` sounds overly complex to me.

robrix:
RAC & [Rx](http://reactivex.io/documentation/operators/materialize-dematerialize.html). both use `materialize`/`dematerialize`

neilpa:
> RAC & Rx. both use materialize/dematerialize

Which is why I suggested it.

gfontenot:
:+1: Updated.

PR #60, opened 2015-06-11
Add function version of `materialize`

ishkawa:
CAUTION: This patch is also not working for now in the same way as #56 .
https://twitter.com/gfontenot/status/608774708159324161

#56 is great, but function version of `materialize` is easier to derive `Result` from failable closure.

```swift
let data = NSData()
let result = materialize {
	try NSJSONSerialization.JSONObjectWithData(data, options: [])
}
```

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -181,12 +181,12 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 
 // MARK: - Derive result from failable closure
 
-public func materialize<T>(f: () throws -> T) -> Result<T, ErrorType> {
-	do {
-		return .success(try f())
-	} catch {
-		return .failure(error)
-	}
+public func materialize<T, E: ErrorType>(f: () throws -> T) -> Result<T, E> {
+    do {
+        return .success(try f())
+    } catch {
+        return .failure(error as! E)
+    }
 }
 
 // MARK: - Cocoa API conveniences
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -181,12 +181,12 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 
 // MARK: - Derive result from failable closure
 
-public func materialize<T, E: ErrorType>(f: () throws -> T) -> Result<T, E> {
-    do {
-        return .success(try f())
-    } catch {
-        return .failure(error as! E)
-    }
+public func materialize<T>(f: () throws -> T) -> Result<T, ErrorType> {
+	do {
+		return .success(try f())
+	} catch {
+		return .failure(error)
+	}
 }
 
 // MARK: - Cocoa API conveniences
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -179,6 +179,15 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 	return left.recoverWith(right())
 }
 
+// MARK: - Derive result from failable closure
+
+public func materialize<T>(f: () throws -> T) -> Result<T, ErrorType> {
+	do {
+		return .success(try f())
+	} catch {
+		return .failure(error)
+	}
+}
 
 // MARK: - Cocoa API conveniences
 


karupanerura:
:+1:

robrix:
Could this be a `static` function instead please?

gfontenot:
We should _probably_ either hold off on merging this, or comment out the implementation till the next beta. I was already planning on doing that for the original `materialize` implementation. It would be nice if `master` could compile.

robrix:
Agreed with @gfontenot.

robrix:
That said, thank you for the pull request, @ishkawa!

Result/Result.swift
```diff
@@ -179,6 +179,15 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 	return left.recoverWith(right())
 }
 
+// MARK: - Derive result from failable closure
+
+public func materialize<T>(f: () throws -> T) -> Result<T, ErrorType> {
```
JaviSoto: I don't know if this would work, but maybe we can provide another function for when the caller knows the type of error being thrown? Something like:

```swift
public func materialize<T, E: ErrorType>(f: () throws -> T) -> Result<T, E> {
	do {
		return .success(try f())
	} catch {
		return .failure(error as E)
	}
}
```

robrix: It _would_ be really nice to retain type info for the errors. I think we could just include this version, rather than overloading.

JaviSoto: *this version* meaning the one I proposed or the one in the PR?

robrix: @JaviSoto: the version you propose here.

ishkawa: It's better to keep type info, but how is `E` inferred? Currently, we can't know what kind of error `() throws -> T` throws. And if we can't get `E` from `() throws -> T`, type annotation is required to call this function as below:

```swift
let data = NSData()
let result: Result<AnyObject, NSError> = materialize {
    try NSJSONSerialization.JSONObjectWithData(data, options: [])
}
```

neilpa: > It's better to keep type info, but how is E inferred?

I haven't experimented with this case, but generally the type-checker can infer from the surrounding code without explicit specification at the call-site.

> Currently, we can't know what kind of error () throws -> T throws.

I suspect there are a lot of cases where you "know" the error, it just can't be documented in the throws clause. And in the cases where you don't, you can use `ErrorType` directly.

ishkawa: My concern is `materialize` function can't be called without type annotation.

If error type can't be documented in throws clause, there is no place to specify `E` except call-site. That is, type of `let result` can't be inferred from return value of `materialized` function. Consequently, type annotation (`Result<AnyObject, NSError>` in my code above) is required to call `materialized`.

neilpa: IMO, explicit specification is better than the alternative of having to later unwrap and cast the contained error. I also expect it will be common to do `return materalize(foo)` where the type info can be inferred from the function declaration where it's required.

ishkawa: OK, that's still acceptable for me. Thank you for your explanation.

The next issue is casting error to `E`. Because it's failable, we should instantiate `E` somehow if casting fails. Is there any way to instantiate `E` in this case?

```swift
public func materialize<T, E: ErrorType>(f: () throws -> T) -> Result<T, E> {
    do {
        return .success(try f())
    } catch let error as E {
		return .failure(error)
	} catch {
		// what should we return here?
	}
}
```

neilpa: We should just go with what @JaviSoto suggested above with the forced cast in the default catch block.

ishkawa: I've edited PR as @JaviSoto suggested, and added forced casting.

Result/Result.swift
```diff
@@ -179,6 +179,15 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 	return left.recoverWith(right())
 }
 
+// MARK: - Derive result from failable closure
+
+public func materialize<T, E: ErrorType>(f: () throws -> T) -> Result<T, E> {
+    do {
+        return .success(try f())
+    } catch {
+        return .failure(error as! E)
```
JaviSoto: Is `as!` (forced) required? For some reason you can always do `errorType as NSError` and that always works :-/
If this is going to be a forced-casting, perhaps I would provide a separate method that just gives you an `ErrorType`?

neilpa: Yea, the force cast is required. The `as NSError` works because `ErrorType` actually requires `_domain` and `_code` properties that are used to fill out the `NSError`. In the case of enums, the compiler automatically generates values for said properties.

jspahrsummers:
I'm pretty nervous about adding a function that can _implicitly_ perform an _explicitly_ unsafe operation. Simply through type inference, you might introduce a crash.

Can `materialize` take the expected type of the error as an argument instead (a la RAC's `promoteErrors`), or something like that?

ishkawa:
> I'm pretty nervous about adding a function that can implicitly perform an explicitly unsafe operation. Simply through type inference, you might introduce a crash.

Absolutely. I've just reverted 6cdd75ce19deab85012b381846aea040c89fe800.

> Can materialize take the expected type of the error as an argument instead (a la RAC's promoteErrors), or something like that?

In either case,  casting is required if error type is passed from call-site.

Summarizing the discussion here:

- 9fdc03197ddec6d82c90e72c924972dd0cd9bfd6 is safe but loses error type info.
- 6cdd75ce19deab85012b381846aea040c89fe800 is unsafe but can keep error type info.

Personally, I think 9fdc03197ddec6d82c90e72c924972dd0cd9bfd6 is better until error type can be  inferred from `() throws -> Void` in the future version of Swift. If `throws` can declare error type like below, both issues are resolved.

```swift
// not work in current version of Swift
public func materialize<T, E: ErrorType>(f: () throws E -> T) -> Result<T, E> {
    do {
        return .success(try f())
    } catch {
        return .failure(error)
    }
}
```

jspahrsummers:
I'm :+1: on this PR as it is currently, but I'll defer to @robrix.

robrix:
Yeah, agreed. Sorry for the delay!

robrix:
Thanks very much @ishkawa! :bow:

ishkawa:
:ok_hand:

ishkawa:
Ah, this PR introduces crash of compiler the same as #56 for now. I'll submit a PR like #61 that comments out `materialize` function to avoid crash of compiler.

PR #63, opened 2015-06-12
Result-watchOS target against the watchOS platform

JaviSoto:
Much like iOS and MacOSX, watchOS needs its own target... I'm currently going through all the dependencies that I use on WatchChess and adding this :laughing:


robrix:
:boom:

PR #64, opened 2015-06-13
Remove VALID_ARCHS for watchOS target so that it uses the default armv7k

JaviSoto:
Sorry I didn't get it right the first time :(


robrix:
It happens, don’t sweat it.

robrix:
Thanks for the PR, @JaviSoto!

PR #66, opened 2015-06-16
Remove commented out `materialize` and replace it with a Result constructor

sync:
- had to Comment out `testErrorsIncludeTheSourceFile` and `testErrorsIncludeTheCallingFunction ` due to compiler segfault
- minor iOS tests target cleanup suppressing the following error: `ld: warning: directory not found for option '-F/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.0.sdk/Developer/Library/Frameworks'`

### DIFF
--- a/ResultTests/ResultTests.swift
+++ b/ResultTests/ResultTests.swift
@@ -22,7 +22,7 @@ final class ResultTests: XCTestCase {
 
 	func testErrorsIncludeTheSourceFile() {
 		let file = __FILE__
-		XCTAssertEqual(Result<(), NSError>.error().file ?? "", file)
+		XCTAssert(Result<(), NSError>.error().file == file)
 	}
 
 	func testErrorsIncludeTheSourceLine() {
@@ -32,9 +32,20 @@ final class ResultTests: XCTestCase {
 
 	func testErrorsIncludeTheCallingFunction() {
 		let function = __FUNCTION__
-		XCTAssertEqual(Result<(), NSError>.error().function ?? "", function)
+		XCTAssert(Result<(), NSError>.error().function == function)
 	}
 
+	// MARK: Try - Catch
+	
+	func testTryCatchProducesSuccesses() {
+		let result: Result<String, NSError> = Result(try tryIsSuccess("success"))
+		XCTAssert(result == success)
+	}
+	
+	func testTryCatchProducesFailures() {
+		let result: Result<String, NSError> = Result(try tryIsSuccess(nil))
+		XCTAssert(result.error == error)
+	}
 
 	// MARK: Cocoa API idioms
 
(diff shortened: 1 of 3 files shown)


robrix:
> minor iOS tests target cleanup

:+1: thank you.

ResultTests/ResultTests.swift
```diff
@@ -31,10 +31,21 @@ final class ResultTests: XCTestCase {
 	}
 
 	func testErrorsIncludeTheCallingFunction() {
-		let function = __FUNCTION__
-		XCTAssertEqual(Result<(), NSError>.error().function ?? "", function)
+//		let function = __FUNCTION__
+//		XCTAssertEqual(Result<(), NSError>.error().function ?? "", function)
```
robrix: Do these tests segfault because of the usage of `Result` or because of the `XCTAssertEqual`? I’d like to be cautious that we’re not just moving a segfault from the API to API consumers.

sync: it segfault because of the usage of `.file` and `.function` so I guess it's :-1:

robrix: Do you know if there’s a radar for this?

sync: I was planning on submitting one this week end, I'll post the radar number here

sync: funny turn out it can get it to crash only when inside the tests target, if same code is added to result itself:

```
func test() {
	let function = __FUNCTION__
	let expected = Result<(), NSError>.error().function ?? ""
	if function == expected {
		
	}
}
```

It doesn't crash.

sync: its `?? ""` causing the crash

sync:
@robrix thanks to c615c0a4df094bafd435d6d6dcd1b12b3a052fe2, compiler is no longer crashing, all tests are passing, also rebased branch on top of master

robrix:
/cc @antitypical/result — I think it’s more idiomatic using `init`, but the semantics are somewhat different—`materialize` takes `f : T throws -> U` and returns a materializing function, whereas this takes `f : () throws -> T`.

Should this replace `materialize` or augment it? One can recover `materialize` by closure, something like so:

```swift
{ f in { x in Result { f(x) } } }
```

sync:
now that we can reintroduce `@autocolsure`, see 6db2b401891eec98611a22293a7cc3cd5c581a7b I definitely find the `init` more idiomatic. I would suggest replacing `materialize`, as well we can use the init method now. `Result(try tryIsSuccess("success"))` is very easy / convenient to use

sync:
this branch is becoming a "fix all the things", but it seems that the latest `ResultType` addition broke the watchOS target, see 22907f335055624ac889c1d4039a730e08b55bf5

robrix:
@sync In the interests of getting this merged sooner, let’s keep everything as you’ve got it here except to restore the commented out `materialize` function. We can always remove it later if we decide it’s entirely redundant.

sync:
doing this right away

sync:
@robrix have a look, once more rebased on top of master, collapsed some of the previous commits

robrix:
:+1: looks good!

robrix:
Thanks very much for the PR, @sync!

PR #67, opened 2015-06-16
Add ResultType protocol to facilitate protocol extension constraints

Thomvis:
Protocols are really powerful in Swift and with Swift 2.0 even more so. By adding the `ResultType` protocol, we can use the type in protocol extensions and in generic constraints.

The primary motivation for this PR is a refactoring of BrightFutures (which has Result as a dependency) that I am working on. With the help of the `ResultType` protocol, I can define a simple `Deferred<T>` type and add methods (such as `onSuccess` and `onFailure`) to it when `T: ResultType`. 
- Relevant code in BrightFutures: https://gist.github.com/Thomvis/0684c36159d173396297ub.com/Thomvis/BrightFutures/blob/swiftier-2.0/BrightFutures/Future.swift#L112
- Concise example:  https://gist.github.com/Thomvis/0684c36159d173396297

The contents of the protocol is debatable, but the current contents are the parts that I found helpful.

:star2: Let me know what you think!

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -131,6 +131,21 @@ public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStr
 	}
 }
 
+public protocol ResultType {
+	typealias Value
+	typealias Error: ErrorType
+	
+	var value: Value? { get }
+	var error: Error? { get }
+	
+	init(value: Value)
+	init(error: Error)
+	
+	func analysis<U>(@noescape ifSuccess ifSuccess: Value -> U, @noescape ifFailure: Error -> U) -> U
+}
+
+extension Result: ResultType { }
+
 
 /// Returns `true` if `left` and `right` are both `Success`es and their values are equal, or if `left` and `right` are both `Failure`s and their errors are equal.
 public func == <T: Equatable, Error: Equatable> (left: Result<T, Error>, right: Result<T, Error>) -> Bool {
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -1,7 +1,7 @@
 //  Copyright (c) 2015 Rob Rix. All rights reserved.
 
 /// An enum representing either a failure with an explanatory error, or a success with a result value.
-public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStringConvertible {
+public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, CustomDebugStringConvertible {
 	case Success(T)
 	case Failure(Error)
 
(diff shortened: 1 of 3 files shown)


Result/Result.swift
```diff
@@ -131,6 +131,21 @@ public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStr
 	}
 }
 
+public protocol ResultType {
```
robrix: Could you add the protocol to a new file, please?

Result/Result.swift
```diff
@@ -131,6 +131,21 @@ public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStr
...
+	var value: Value? { get }
+	var error: Error? { get }
+	
+	init(value: Value)
+	init(error: Error)
+	
+	func analysis<U>(@noescape ifSuccess ifSuccess: Value -> U, @noescape ifFailure: Error -> U) -> U
+}
+
+extension Result: ResultType { }
```
robrix: Could you please declare this on `Result` rather than in an extension?

Result/Result.swift
robrix: Also, simple documentation comments for the protocol and its members would be greatly appreciated :bow:

robrix:
Thanks for the PR, @Thomvis! @antitypical/result, I’d love to know what you think. I’m quite :+1: on what this enables for APIs using `Result`.

mdiep:
I'm sad that a protocol is required to do this, but very :+1: on what it enables.

Thomvis:
Thanks for the feedback. I think they're all valid and will follow up over the weekend.

Result/Result.swift
```diff
@@ -131,6 +131,21 @@ public enum Result<T, Error: ErrorType>: CustomStringConvertible, CustomDebugStr
 	}
 }
 
+public protocol ResultType {
+	typealias Value
+	typealias Error: ErrorType
+	
+	var value: Value? { get }
+	var error: Error? { get }
```
robrix: Given `analysis`, you can provide `value` and `error` in a protocol extension. That reduces the requirements to two introduction rules (the `init`s) and one elimination rule (`analysis`) covering both alternatives, which is its logical minimum :tada:

Thomvis:
I've updated the PR according to your feedback. I'm not very confident about the documentation, so feel free to give feedback or change it if you decide to merge it in.

robrix:
Needs a merge from `master` :pray:

Result/ResultType.swift
```diff
@@ -0,0 +1,31 @@
+//  Copyright (c) 2015 Rob Rix. All rights reserved.
+
+/// A type that can represent either failure with an error or success with a result value.
+public protocol ResultType {
+	typealias Value
+	typealias Error: ErrorType
+	
+	/// Constructs a succesful result wrapping a `value`.
```
robrix: Nitpick: `successful`

robrix:
Docs are good, modulo the one typo; thanks so much @Thomvis!

Thomvis:
:white_check_mark: corrected typo
:white_check_mark: merged from `master`

neilpa:
:sparkling_heart: the direction here

robrix:
It is a delight to press this merge button! :tada:

PR #68, opened 2015-06-20
Comment out materialize function

ishkawa:
The same as #61. See https://github.com/antitypical/Result/pull/60#issuecomment-113681768 for more detail :bow:


robrix:
Thanks @ishkawa.

PR #69, opened 2015-06-24
Remove success and failure static methods

ikesyo:
Enum cases with associated values can now be used as functions from Xcode 7 beta 2.


robrix:
Thanks @ikesyo! Could you please also change the (commented out) uses in `materialize`?

ikesyo:
Thanks @robrix, I've overlooked it. :sweat_smile: Fixed!

robrix:
:bow: thank you so much for the PR!

PR #78, opened 2015-09-02
Explicitly set code signing

gfontenot:
Not having this set was causing intermittent build failures when building with Carthage.

Similar to ReactiveCocoa/ReactiveCocoa#2292

I'm not sure that this is the _best_ solution to this, but I'm not sure it's a bad solution either. If someone has more knowledge behind why this is happening, I'm all ears.

### DIFF
--- a/Result.xcodeproj/project.pbxproj
+++ b/Result.xcodeproj/project.pbxproj
@@ -477,6 +477,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
 				COMBINE_HIDPI_IMAGES = YES;
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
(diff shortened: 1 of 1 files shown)


JaviSoto:
:+1:

Result.xcodeproj/project.pbxproj
```diff
@@ -716,6 +718,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
```
neilpa: Probably want to delete these though

Result.xcodeproj/project.pbxproj
```diff
@@ -740,6 +743,7 @@
...
```
neilpa: And here

gfontenot:
@neilpa good catch, fixed those.

neilpa:
Seeing as I made the same change for RAC :+1: as well.

PR #79, opened 2015-09-02
Update materialize functions

gfontenot:
- Enable free materialize function and use NSError:

That bug isn't fixed yet, but this also won't ever compile. Somewhat
Reasonably, you are unable to use a protocol as a concrete type for a
generic type that has a type constraint with that protocol.

To get around this, we should use `NSError` for the error type, since we
can always safely cast from `ErrorType` to `NSError`.

Not sure if we should make this change to the initializer as well. Seems like
it might be safer than using a force cast. 

- Remove static `materialize` function:

Now that we have the initializer that takes a function, this seems less
interesting. Returning a function is much less useful than just passing a
function and getting a Result.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -158,14 +158,13 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 
 // MARK: - Derive result from failable closure
 
-// Disable until http://www.openradar.me/21341337 is fixed.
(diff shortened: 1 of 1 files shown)


gfontenot:
ping @antitypical/result
Seems like this might be good to make a decision about before releasing Swift 2.0 support.

robrix:
:+1: from me.

PR #81, opened 2015-09-15
Add simple Podspec to the repo

### LINKED ISSUE #73: Add a podspec (supporting watchOS, &c)
please update the platforms in the pod spec to support watchOS for 0.6-beta.1

gfontenot:
Fixes #73

### DIFF
--- a/Result.podspec
+++ b/Result.podspec
@@ -4,11 +4,12 @@ Pod::Spec.new do |s|
   s.summary      = 'Swift type modelling the success/failure of arbitrary operations'
 
   s.homepage     = 'https://github.com/antitypical/Result'
-  s.license      = 'MIT'
+  s.license      = { :type => 'MIT', :file => 'LICENSE' }
   s.author       = { 'Rob Rix' => 'rob.rix@github.com' }
-  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version.to_s }
+  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version }
   s.source_files  = 'Result/*.swift'
   s.requires_arc = true
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
+  s.watchos.deployment_target = '2.0'
 end
--- a/Result.podspec
+++ b/Result.podspec
@@ -1,0 +1,14 @@
+Pod::Spec.new do |s|
+  s.name         = 'Result'
+  s.version      = '0.6.0-beta.2'
+  s.summary      = 'Swift type modelling the success/failure of arbitrary operations'
+
+  s.homepage     = 'https://github.com/antitypical/Result'
+  s.license      = 'MIT'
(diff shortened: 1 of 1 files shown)


Result.podspec
```diff
@@ -0,0 +1,14 @@
...
+  s.summary      = 'Swift type modelling the success/failure of arbitrary operations'
+
+  s.homepage     = 'https://github.com/antitypical/Result'
+  s.license      = 'MIT'
+  s.author       = { 'Rob Rix' => 'rob.rix@github.com' }
+  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version.to_s }
+  s.source_files  = 'Result/*.swift'
+  s.requires_arc = true
+  s.ios.deployment_target = '8.0'
+  s.osx.deployment_target = '10.9'
```
NachoSoto: I think you'll need deployment targets for `watchOS` and `tvOS` too?

kylef: I don't believe so, since their default values are watchOS 2.0 and tvOS 9.0.

Result.podspec
```diff
@@ -0,0 +1,14 @@
+Pod::Spec.new do |s|
+  s.name         = 'Result'
+  s.version      = '0.6.0-beta.2'
+  s.summary      = 'Swift type modelling the success/failure of arbitrary operations'
+
+  s.homepage     = 'https://github.com/antitypical/Result'
+  s.license      = 'MIT'
```
kylef:
```ruby
{ :type => 'MIT', :file => 'LICENSE' }
```

Result.podspec
```diff
@@ -0,0 +1,14 @@
...
+  s.author       = { 'Rob Rix' => 'rob.rix@github.com' }
+  s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version.to_s }
```
kylef: The `to_s` on `s.version` is unnecessary.

Result.podspec
ikesyo: As described in the post, we must explicitly set it: http://blog.cocoapods.org/CocoaPods-0.38/

> What this means in practice, is that not all Pods are automatically available by default. Any Pod that can support watchOS needs to explicitly add support by adding s.watchos.deployment_target. It is an entirely new platform.

NachoSoto: Thanks @ikesyo! I thought I read that somewhere.

gfontenot: Good catch.

Result.podspec
gfontenot: derp

Result.podspec
gfontenot: Added the watchOS target. Can't add the tvOS target yet as it isn't actually supported by CocoaPods.

gfontenot:
I'll merge this today unless there are any real objections.

ashfurrow:
Looks fab :tada:

robrix:
@gfontenot: :heart:

PR #83, opened 2015-09-16
Remove tvOS target from master

gfontenot:
We should release 0.6 with Swift 2.0, but `tvOS` is only included in Xcode 7.1,
which in beta right now. To fix this, we should remove the `tvOS` target from
`master` until we've released version 0.6, and start working on version 0.7 to
add the `tvOS` target.

Fixes #83


gfontenot:
Not sure how we want to handle the git history here. I have a `tvos` branch created. Should that one be the same as `master` but with a revert of this revert?

NachoSoto:
Yeah that seems fine. I'd recommend leaving the target though! That way it'll be easier to integrate back and you won't have to deal with project conflicts. You can just remove the scheme :)

NachoSoto:
(Related: https://openradar.appspot.com/2264503)

gfontenot:
That's actually a much better idea than doing _any_ reverts. I'll update this PR to remove the scheme, and will re-add it to the `tvos` branch.

NachoSoto:
Awesome :)

PR #86, opened 2015-09-22
Move map, flatMap, and &&& to ResultType

bencochran:
As it stand, these can all be implemented in terms of `ResultType`. And doing so opens up the possibilities for more things, like this:

```swift
extension SequenceType where Generator.Element: ResultType {
    /// Flattens a sequence of Results
    ///
    /// Returns a single Result by combining the values of all `Success`es in this sequence, or re-wrapping the first encountered `Failure`
    func flatten() -> Result<[Generator.Element.Value], Generator.Element.Error> {
        return reduce(.Success([])) { result, element in
            return (result &&& element).map { $0 + [$1] }
        }
    }
}
```

(which isn’t part of this pull request, but could be if there’s interest)

### DIFF
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -35,7 +35,7 @@ public extension ResultType {
 	}
 
 	/// Returns the result of applying `transform` to `Success`es’ values, or re-wrapping `Failure`’s errors.
-	func flatMap<U>(@noescape transform: Value -> Result<U, Error>) -> Result<U, Error> {
+	public func flatMap<U>(@noescape transform: Value -> Result<U, Error>) -> Result<U, Error> {
 		return analysis(
 			ifSuccess: transform,
 			ifFailure: Result<U, Error>.Failure)
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -45,11 +45,11 @@ public extension ResultType {
 // MARK: - Operators
 
 infix operator &&& {
-/// Same associativity as &&.
-associativity left
+	/// Same associativity as &&.
+	associativity left
 
-/// Same precedence as &&.
-precedence 120
+	/// Same precedence as &&.
+	precedence 120
 }
 
 /// Returns a Result with a tuple of `left` and `right` values if both are `Success`es, or re-wrapping the error of the earlier `Failure`.
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -28,4 +28,16 @@ public extension ResultType {
 	var error: Error? {
 		return analysis(ifSuccess: { _ in nil }, ifFailure: { $0 })
 	}
+
+	/// Returns a new Result by mapping `Success`es’ values using `transform`, or re-wrapping `Failure`s’ errors.
+	public func map<U>(@noescape transform: Value -> U) -> Result<U, Error> {
+		return flatMap { .Success(transform($0)) }
+	}
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or re-wrapping `Failure`’s errors.
+	func flatMap<U>(@noescape transform: Value -> Result<U, Error>) -> Result<U, Error> {
+		return analysis(
+			ifSuccess: transform,
+			ifFailure: Result<U, Error>.Failure)
+	}
 }
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -41,3 +41,18 @@ public extension ResultType {
 			ifFailure: Result<U, Error>.Failure)
 	}
 }
+
+// MARK: - Operators
+
+infix operator &&& {
+/// Same associativity as &&.
+associativity left
+
+/// Same precedence as &&.
+precedence 120
+}
+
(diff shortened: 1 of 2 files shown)


Result/ResultType.swift
```diff
@@ -28,4 +28,31 @@ public extension ResultType {
 	var error: Error? {
 		return analysis(ifSuccess: { _ in nil }, ifFailure: { $0 })
 	}
-}
\ No newline at end of file
+
+	/// Returns a new Result by mapping `Success`es’ values using `transform`, or re-wrapping `Failure`s’ errors.
+	public func map<U>(@noescape transform: Value -> U) -> Result<U, Error> {
+		return flatMap { .Success(transform($0)) }
+	}
```
robrix: I have philosophical difficulties with a method on `ResultType` producing a `Result`. Unfortunately, we can’t make it return `Self`-parameterized-by-`U` (because Swift lacks higher-ranked types). That’s why I made this a method on `Result` in the first place; presumably if you’re `map`ing some other `ResultType`, then you expect to get an instance of that type back.

(I have the same objection to `SequenceType.map` returning `Array`, but as yet no one has put me in charge of the Swift stdlib.)

robrix:
Although I’m not in favour of producing `Result` from `map` and such, I would (in theory) be amenable to a `ResultConvertible` protocol:

```swift
protocol ResultConvertible {
	typealias Value
	typealias Error: ErrorType
	init(result: Result<Value, Error>)
	var result: Result<Value, Error>
}
```

We could additionally make `ResultType` conform to it, with default implementations:

```swift
protocol ResultType: ResultConvertible {
	…
}

extension ResultType {
	init(result: Result<Value, Error>) {
		switch result {
		case let .Success(s):
			self.init(value: s)
		case let .Failure(e):
			self.init(error: e)
		}
	}

	var result: Result<Value, Error> {
		return analysis(ifSuccess: Result.Success, ifFailure: Result.Failure)
	}
}
```

This would be sufficient to recover the `SequenceType.flatten` implementation you give, without loss of generality to `map` &c.

neilpa:
That's an interesting approach I hadn't considered before to workaround the limitations of protocol extensions.

Thomvis:
I think having map etc. on `ResultType` is a good idea. It would be even better if we could return Self with new type parameters. (Swift 3 maybe?) The workaround is indeed interesting, but if I understand it correctly, it would require to get `result` from a `map`'s result (i.e. `res.map(f).result`), right?

Related: I like the `.flatten` operation you propose, but is 'flatten' the right name? I've been using that exact operation, but have been calling it 'sequence' (after [Scala](http://www.scala-lang.org/api/current/index.html#scala.concurrent.Future$@sequence[A,M[X]<:TraversableOnce[X]](in:M[scala.concurrent.Future[A]])(implicitcbf:scala.collection.generic.CanBuildFrom[M[scala.concurrent.Future[A]],A,M[A]],implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[M[A]]))

robrix:
@neilpa: Thank you. I’ve been using it a lot lately.

@Thomvis: That’s correct; there’s a little extra wrapping/unwrapping necessary for now.

My main concern here is that when Swift supports higher-rank types (which I’m assuming it will—maybe that’s not a reasonable assumption, though) I would very much want `ResultType.map` to return `Self`. I suppose we would likely be releasing a new major version at that time anyway, however, and thus could change the API at that time. What do we think about viewing this PR as an interim solution?

There’s also the question of how many uses of `ResultType` even exist—I suspect there are vanishingly few.

Result/ResultType.swift
```diff
@@ -28,4 +28,31 @@ public extension ResultType {
...
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or re-wrapping `Failure`’s errors.
+	func flatMap<U>(@noescape transform: Value -> Result<U, Error>) -> Result<U, Error> {
```
robrix: Orthogonal to the more philosophical discussion, could you please mark `flatMap` as `public`?

bencochran:
- [x] Make `flatMap` public.

I’m slightly confused about the `ResultConvertible` proposal though. In @Thomvis’s example he used `res.map(f).result` but wouldn’t it be `res.result.map(f)`? That is, is the proposal to move `map` and `flatMap` back to `Result` and, in the case of my `Array` extension, first convert the `ResultType` to a `Result` to perform my operations on?

As for the philosophical discussion. My general thinking is that *some* limitations of the language are going to come through in API design (as in your example of `SequenceType.map`). Given that higher-ranked types will(/would) change the shape of the language and stdlib enough, I suspect there will be much code to revisit in general (another Swift 1.2→2-like incompatibility at worst, old libraries feeling old at best).

robrix:
@bencochran: Yeah, I think that was just the usual “human brains are no substitute for a good typechecker” sort of mixup :blush:

Overall, I agree. We can revisit this when we get higher-rank types.

Unfortunately, merging #89 introduced conflict(s). I think we can ship this once they’re resolved.

bencochran:
Aha! Ok, glad to know I wasn’t missing something.

Anywho, this should cleanly merge now.

robrix:
:+1: Thanks for the PR, @bencochran!

PR #89, opened 2015-09-23
ResultType extension properties should be public

robrix:
`ResultType`’s `value` & `error` properties are `public`, and can be overridden by conforming types.


Result/ResultType.swift
```diff
@@ -20,12 +20,12 @@ public protocol ResultType {
...
 	
 	/// Returns the value if self represents a success, `nil` otherwise.
-	var value: Value? {
+	public var value: Value? {
 		return analysis(ifSuccess: { $0 }, ifFailure: { _ in nil })
 	}
 	
 	/// Returns the error if self represents a failure, `nil` otherwise.
-	var error: Error? {
+	public var error: Error? {
```
robrix: I’ve been wondering if these should be declared in `ResultType` as well. Default implementations would be provided by this extension, but declaring them in `ResultType` would allow conforming types to provide their own implementations, which could be useful for efficiency or debugging.

neilpa: :+1: Unless there's a good reason to prohibit custom implementations, I think it makes sense to also declare default implementations as required by the protocol.

robrix: :ok:, it is done.

neilpa:
Cool, I'd say ship it.

PR #92, opened 2015-10-03
Fix code of fixture error for 32-bit devices

ishkawa:
Replaces `0xdeadbeef` in test code with `0x01234567`.
`0xdeadbeef ` causes following build error on 32-bit devices.

```
ResultTests/ResultTests.swift:101:61: error: integer literal '3735928559' overflows when stored into 'Int'
let error = NSError(domain: "com.antitypical.Result", code: 0xdeadbeef, userInfo: nil)
```

See https://travis-ci.org/ishkawa/Result/builds/83435179 for the full content of xcodebuild log.

### DIFF
--- a/ResultTests/ResultTests.swift
+++ b/ResultTests/ResultTests.swift
@@ -98,8 +98,8 @@ final class ResultTests: XCTestCase {
 // MARK: - Fixtures
 
 let success = Result<String, NSError>.Success("success")
-let error = NSError(domain: "com.antitypical.Result", code: 0x01234567, userInfo: nil)
-let error2 = NSError(domain: "com.antitypical.Result", code: 0x12345678, userInfo: nil)
+let error = NSError(domain: "com.antitypical.Result", code: 1, userInfo: nil)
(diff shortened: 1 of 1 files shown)


robrix:
Thanks for the PR!

Since this makes `error` and `error2` the same, I wonder if we could just use a single error, or alternatively use different codes (e.g. `1` and `2`) while still avoiding overflow on 32-bit.

ishkawa:
> this makes error and error2 the same

No, there're different each other, `0x01234567 ` and `0x12345678 `. Actually, there're confusing, so I have replaced them with `1` and `2`!

robrix:
Haha, I clearly hadn’t caught that :smile: Nice one!

This is very good, thank you @ishkawa!

ishkawa:
:smile:

robrix:
@ishkawa: I’ve added you as a collaborator. We prefer a PR & review workflow for changes, as usual, but you can feel free to review, tag, etc. now as well :smile:

ishkawa:
Glad to hear that! Thank you :sparkles:

PR #95, opened 2015-10-06
Add autoclosure attribute to materialize() as with init()

ikesyo:


ikesyo:
This can be used as follows:

```
// With this
materialize(try NSFileManager.defaultManager().removeItemAtURL(someURL))
// Without this
materialize { try NSFileManager.defaultManager().removeItemAtURL(someURL) }
```

robrix:
This seems reasonable to me, although I would note that it’s a breaking API change.

@antitypical/result: Any thoughts?

gfontenot:
I'm :+1: on API consistency.

robrix:
That works for me :+1:

ikesyo:
Thanks! :v:

robrix:
Thank _you_!

PR #97, opened 2015-10-08
Don't code sign builds for the iOS simulator

gfontenot:
There's no real reason to do this, and in fact it breaks CI builds using
Carthage that aren't ever going to try to build for a device anyway.

### DIFF
--- a/Result.xcodeproj/project.pbxproj
+++ b/Result.xcodeproj/project.pbxproj
@@ -913,6 +913,7 @@
 			buildSettings = {
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
+				"CODE_SIGN_IDENTITY[sdk=iphonesimulator*]" = "";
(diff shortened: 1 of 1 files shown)


tonyd256:
If this does the trick then :+1:

JaviSoto:
:+1: I think this is right

gfontenot:
@JaviSoto any thoughts on the watchOS stuff? I know you've dealt with this more than I have. I'm assuming that (and the tvOS target) need this same fix, but I'm not sure if I should be choosing the iOS or watchOS simulators.

JaviSoto:
Yes, I believe the set up should be the same for those. Just with `watchsimulator` and `tvsimulator` (not sure about this last value)

gfontenot:
Cool, I'll start with that then. Updating this as soon as the Xcode beta verifies :disappointed:

gfontenot:
Turns out: `appletvsimulator` is the name of the tvOS simulators. Good job with consistency, Apple!

PR #100, opened 2015-10-14
Add mapError() & flatMapError()

inamiy:
Hi. I just wanted an Error type conversion, so I will send this pull request.

### DIFF
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -50,6 +50,18 @@ public extension ResultType {
 			ifSuccess: transform,
 			ifFailure: Result<U, Error>.Failure)
 	}
+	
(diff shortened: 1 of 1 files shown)


robrix:
Thanks for the pull request!

I’m :+1: to the implementations, but I’m curious what others in @antitypical/result think about extending the API thus. Is this a common enough use case to be worth including in the framework itself?

mdiep:
:+1:

Thomvis:
I think it can definitely be useful e.g. when converting from NSError to your own error type or when using third party libraries that use Result in their API with their own error type. :+1: from me as well!

robrix:
Works for me :+1:

inamiy:
Thanks for merge :smiley:

PR #104, opened 2015-10-22
Disable bitcode for Mac target for Xcode 7.1 compatibility

ikesyo:
If bitcode is enabled for Mac target, the build would fail. Additionaly `ENABLE_BITCODE` set to YES only for iOS target, the setting would not be appeared in Xcode 7.1 for watchOS and tvOS target anymore (since they must be bitcode-enabled?).


robrix:
:+1:

robrix:
Thanks very much for the PR, @ikesyo!

PR #105, opened 2015-10-22
Add <^> (fmap) & <*> (ap) for applicative style support.

inamiy:
It will be nice if we can add Haskell's `fmap` and `ap` to support applicative style
as https://github.com/thoughtbot/Argo 's `Decoded<T>` type can do.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -181,6 +181,26 @@ public func `try`(function: String = __FUNCTION__, file: String = __FILE__, line
 
 // MARK: - Operators
 
+infix operator <^> {
+	associativity left
+	precedence 140
+}
+
+/// fmap
+public func <^> <T, U, Error> (@noescape transform: T -> U, result: Result<T, Error>) -> Result<U, Error> {
+	return result.map(transform)
+}
+
+infix operator <*> {
+	associativity left
+	precedence 140
+}
+
+/// ap
+public func <*> <T, U, Error> (transform: Result<T -> U, Error>, @autoclosure result: () -> Result<T, Error>) -> Result<U, Error> {
+	return transform.flatMap { f in result().map(f) }
+}
+
 infix operator >>- {
 	// Left-associativity so that chaining works like you’d expect, and for consistency with Haskell, Runes, swiftz, etc.
 	associativity left
--- a/ResultTests/ResultTests.swift
+++ b/ResultTests/ResultTests.swift
@@ -92,6 +92,61 @@ final class ResultTests: XCTestCase {
 		let resultFailureRight = success &&& failure2
 		XCTAssert(resultFailureRight.error == error2)
 	}
+	
+	// MARK: Functor, Applicative, Monad
+	
+	func testFunctor() {
+		let result1 = { $0.characters.count } <^> success
+		XCTAssertTrue(result1 == .Success(7))
+		
+		let result2 = { $0.characters.count } <^> failure
+		XCTAssertTrue(result2 == .Failure(error))
+	}
+	
+	func testApplicative() {
+		let result1 = .Success({ $0.characters.count }) <*> success
+		XCTAssertTrue(result1 == .Success(7))
+		
+		let result2 = .Success({ $0.characters.count }) <*> failure
+		XCTAssertTrue(result2 == .Failure(error))
+		
+		let result3: Result<Int, NSError> = .Failure(error) <*> success
+		XCTAssertTrue(result3 == .Failure(error))
+		
+		let result4: Result<Int, NSError> = .Failure(error) <*> failure
+		XCTAssertTrue(result4 == .Failure(error))
+	}
+	
+	func testApplicativeStyle() {
+		let f: Int -> Int -> Int = { x in { y in x + y } }
+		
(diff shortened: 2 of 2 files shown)


Result/Result.swift
```diff
@@ -181,6 +181,26 @@ public func `try`(function: String = __FUNCTION__, file: String = __FILE__, line
...
+public func <^> <T, U, Error> (@noescape transform: T -> U, result: Result<T, Error>) -> Result<U, Error> {
+	return result.map(transform)
+}
+
+infix operator <*> {
+	associativity left
+	precedence 140
+}
+
+/// ap
```
robrix: Could you please document these operators with their use and relationships to other Swift operators?

ResultTests/ResultTests.swift
```diff
@@ -92,6 +92,61 @@ final class ResultTests: XCTestCase {
...
+		let result2 = success >>- { _ in Result<Int, NSError>.Failure(error) }
+		XCTAssertTrue(result2 == .Failure(error))
+		
+		let result3 = failure >>- { .Success($0.characters.count) }
+		XCTAssertTrue(result3 == .Failure(error))
+		
+		let result4 = failure >>- { _ in Result<Int, NSError>.Failure(error) }
+		XCTAssertTrue(result4 == .Failure(error))
+	}
+
```
robrix: :clap: Thank you for the tests! :bow:

Result/Result.swift
```diff
@@ -181,6 +181,26 @@ public func `try`(function: String = __FUNCTION__, file: String = __FILE__, line
...
+
+/// fmap
+public func <^> <T, U, Error> (@noescape transform: T -> U, result: Result<T, Error>) -> Result<U, Error> {
+	return result.map(transform)
+}
+
+infix operator <*> {
+	associativity left
+	precedence 140
+}
```
robrix: Would it be reasonable to add `*>` and `<*` at the same priority (and with the same meanings as in Haskell)?

robrix:
Thank you very much for the PR! :bow:

neilpa:
Wondering out loud if this will be a turnoff to people who've never used Haskell and don't like custom operators. Should we instead expect people that want custom operators to use something like [Runes](https://github.com/thoughtbot/Runes)?

NachoSoto:
I was thinking the same thing.
Also, the fact that there would be 2 different ways of accomplishing the same thing is going to lead to confusion: users will wonder what the difference between using the operator versus the method is

neilpa:
Or [link with runes directly](https://github.com/thoughtbot/Runes#adding-runes-as-a-framework-dependency) if you still think it's valuable. But I'm still :-1: on defining these in `Result` and would prefer to keep things simple.

neilpa:
What I'd rather see is new applicative framework on top of `Result` and some of the other core types like `Either` that links the `Runes.framework`. Seems like a better model for those that want to opt-in to the Haskell style.

robrix:
Hm, these are good points. I’ll bow to majority opinion on this one. @antitypical/result: Anybody else have strong feels?

inamiy:
I'm having a difficulty when I need to handle multiple intermediate results and combine all of them into final one, i.e.:

```
let finalResult = f <$> result1 <*> result2 <*> result3 <*> ...
```

Without applicative style support (using custom operators), I have to write like:
	
```
let finalResult = result1.map { f($0) }.flatMap { result2.map($0) }.flatMap { ...
```

which is quite unreadable and not simple.

Though it's not a bad idea to let other framework implement this extension, it will be a bit tired to import all of those frameworks.
I'm hoping `Result` will work as standalone, just as `>>-` is already implemented previously.

inamiy:
And yes, implementing to other `Either` module will be another solution.

neilpa:
The problem is that for a lot of people these operators are "write once, read never". I'm an admitted Haskell noob but know enough of the basics. However, I haven't found the headspace to retain the meaning behind the infix operators. Whereas `map`, `flatMap`, etc. are very obvious and easy to read, especially with trailing closure syntax.

rnapier:
I agree with @neilpa. I'd rather keep `Result` simple, and let people layer operators on with other frameworks. (I acknowledge that having lots of little frameworks is currently a bit of a pain in Swift, but I'd rather work on improving that than making `Result` bigger. Swift frameworks will get better. Once you add things to `Result`, it's hard to take them out.)

Because operators are defined globally (forcing a particular associativity and precedence for all operands), I consider them a very heavy lift for a framework. We should expect, at a minimum, that the majority of users will want them. I don't believe this is the case for `Result`. In fact, I don't *want* it to be the case. I think devs should be completely comfortable with `Result` even if they have barely a glimmer of understanding of FP. `Result` is the gateway drug after all.

BTW, how is Swift2's compile performance w/ applicative? I've found that most operator-chaining in Swift tends to lead to exponential build times (even just `+` in some cases). I know it's been getting better, but is it now good enough that applicative style doesn't kill your build times?

inamiy:
@rnapier
Oh yeah, long build time was one really bad part of using applicative style in _current_ Swift 2...
Maybe, this pull request was still too early stage for many of us, including compiler :cry:

OK, it seems I should wait for Swift to mature, so I will preserve this branch in salt for a while.
I will close this issue for now.

PR #112, opened 2015-10-30
Add a `materialize()` overload without autoclosure

ikesyo:
We can't use autoclosure one with explicit (trailing) closure. :bow:


robrix:
Thanks! This looks good to me :+1:

PR #113, opened 2015-10-30
[travis] Test tvOS target and build watchOS target

ikesyo:


mdiep:
Why isn't watchOS tested? :confused:

ikesyo:
There is no `XCTest` in watchsimulator, so we can't run it.

mdiep:
Wat.

robrix:
> There is no `XCTest` in watchsimulator, so we can't run it nor build the test target.

Gary Oldman playing Jean-Baptiste Immanuel Zorg in The Fifth Element, laughing hysterically and then crying

PR #114, opened 2015-11-03
Add tryMap. Like flatMap but for Swift2 error handling. 

bwhiteley:
The tryMap function allows code that throws to be more easily wrapped.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -146,7 +146,7 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 
 // MARK: - Derive result from failable closure
 
-public func materialize<T>(f: () throws -> T) -> Result<T, NSError> {
+public func materialize<T>(@noescape f: () throws -> T) -> Result<T, NSError> {
 	return materialize(try f())
 }
 
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -200,5 +200,11 @@ public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T 
 	return result.flatMap(transform)
 }
 
+/// Returns a new Result by mapping `Success`es’ values using `transform`, or wrapping errors thrown as `Failure`s’.
+///
+/// This is a synonym for `tryMap`.
+public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T throws -> U) -> Result<U, Error> {
+    return result.tryMap(transform)
+}
 
 import Foundation
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -51,6 +51,18 @@ public extension ResultType {
 			ifFailure: Result<U, Error>.Failure)
 	}
 	
+	/// Returns a new Result by mapping `Success`es’ values using `transform`, or wrapping errors thrown as `Failure`s’.
+	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+        return flatMap { value in
+            do {
+                return .Success(try transform(value))
+            }
+            catch {
+                return .Failure(error as! Error)
+            }
+        }
+    }
+
 	/// Returns a new Result by mapping `Failure`'s values using `transform`, or re-wrapping `Success`es’ values.
 	public func mapError<Error2>(@noescape transform: Error -> Error2) -> Result<Value, Error2> {
 		return flatMapError { .Failure(transform($0)) }
--- a/ResultTests/ResultTests.swift
+++ b/ResultTests/ResultTests.swift
@@ -89,6 +89,16 @@ final class ResultTests: XCTestCase {
 		XCTAssertNil(result.error)
 	}
 
+	func testTryMapProducesSuccess() {
+		let result = success.tryMap(tryIsSuccess)
+		XCTAssert(result == success)
+	}
+
+	func testTryMapProducesFailure() {
+		let result = Result<String, NSError>.Success("fail").tryMap(tryIsSuccess)
+		XCTAssert(result == failure)
+	}
+
 	// MARK: Operators
 
 	func testConjunctionOperator() {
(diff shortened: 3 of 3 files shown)


Result/Result.swift
```diff
@@ -200,5 +200,11 @@ public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T
 	return result.flatMap(transform)
 }
 
+/// Returns a new Result by mapping `Success`es’ values using `transform`, or wrapping errors thrown as `Failure`s’.
+///
+/// This is a synonym for `tryMap`.
+public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T throws -> U) -> Result<U, Error> {
+    return result.tryMap(transform)
+}
```
robrix: I’m not keen on overloading `>>-` to mean two different things.

Result/ResultType.swift
```diff
@@ -51,6 +51,18 @@ public extension ResultType {
...
+	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+        return flatMap { value in
+            do {
+                return .Success(try transform(value))
+            }
+            catch {
+                return .Failure(error as! Error)
+            }
+        }
+    }
```
robrix: Indentation is a bit weird herein. Tabs, please.

Result/ResultType.swift
```diff
@@ -51,6 +51,18 @@ public extension ResultType {
...
 	}
 	
+	/// Returns a new Result by mapping `Success`es’ values using `transform`, or wrapping errors thrown as `Failure`s’.
+	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+        return flatMap { value in
+            do {
+                return .Success(try transform(value))
+            }
+            catch {
+                return .Failure(error as! Error)
```
robrix: This is about half of what’s wrong with Swift’s error handling semantics in a nutshell 😞

Would it be better to add a constrained extension such that `Error` types which can be constructed with an arbitrary `ErrorType` get `tryMap` and others do not?

Result/Result.swift
```diff
@@ -146,7 +146,7 @@ public func ?? <T, Error> (left: Result<T, Error>, @autoclosure right: () -> Res
 
 // MARK: - Derive result from failable closure
 
-public func materialize<T>(f: () throws -> T) -> Result<T, NSError> {
+public func materialize<T>(@noescape f: () throws -> T) -> Result<T, NSError> {
```
robrix: :+1:

robrix:
Thanks for the PR!

Result/ResultType.swift
bwhiteley: I'll try that. Thanks.

Result/Result.swift
neilpa: Yea, this should be limited to `flatMap`

Result/ResultType.swift
bwhiteley: @robrix Can you provide a hint how to accomplish such a constrained extension? 
Also, how is this worse than the existing initializer? 
public init(@autoclosure _ f: () throws -> T) {
		do {
			self = .Success(try f())
		} catch {
			self = .Failure(error as! Error)
		}
	}

Result/ResultType.swift
```diff
@@ -51,6 +51,18 @@ public extension ResultType {
 			ifFailure: Result<U, Error>.Failure)
 	}
 	
+	/// Returns a new Result by mapping `Success`es’ values using `transform`, or wrapping errors thrown as `Failure`s’.
+	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+        return flatMap { value in
+            do {
+                return .Success(try transform(value))
+            }
+            catch {
```
Thomvis: For consistency I'd say that `catch` should be on the same line as the closing brace of `do`.

Result/ResultType.swift
robrix: Good point, it really isn’t any worse.

robrix:
Maybe it’d be better to change `map` and/or `flatMap` to accommodate `throws` instead. Typically these would be written using `throws` and `rethrows`, but in our case they’d catch instead of rethrowing.

@antitypical/result: Any thoughts?

Result/Result.swift
bwhiteley: @robrix @neilpa is there a different operator you recommend?

robrix: I don’t know of an operator that has the appropriate semantics, and I’d rather not invent one. A method without an operator is totally acceptable.

Result/ResultType.swift
bwhiteley: @robrix How about this? 
In ResultType.swift:
```swift
/// Protocol used to constrain a version of `map` for `transform`s that throw.
public protocol ErrorTypeConvertible: ErrorType {
	static func errorFromErrorType(error:ErrorType) -> ConvertibleType
}

public extension ResultType where Error: ErrorTypeConvertible {
	
	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
	public func map<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
		return flatMap { value in
			do {
				return .Success(try transform(value))
			} catch {
				return .Failure(Error.errorFromErrorType(error) as! Error)
			}
		}
	}
}
```

In Result.swift:
```swift
/// Make NSError conform to ErrorTypeConvertible
extension NSError: ErrorTypeConvertible {
	public static func errorFromErrorType(error: ErrorType) -> NSError {
		return error as NSError
	}
}
```

bwhiteley:
Closing this and opened #118 instead.

Result/Result.swift
neilpa: > A method without an operator is totally acceptable.

:+1:

bwhiteley: @robrix @neilpa  I agree it's good to introduce new operators sparingly. The newly proposed `map` behaves a lot like `flatMap`. The problem is that you can't currently mix and match `>>-` with `flatMap` or the newly proposed `map`. I wonder if this can be fixed by adjusting the precedence of `>>-`.

neilpa: Personally, I'd like to see `>>-` removed and not expose any operators.

PR #118, opened 2015-11-06
Add a version of `map` that converts thrown errors to `Failure`s.

bwhiteley:
Add a variant of `map` that handles `transform`s that throw errors and converts errors to `Failure`s.
This `map` is constrained to `Result`s with `Error` types that can be constructed from arbitrary `ErrorType`s.

### DIFF
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -64,6 +64,26 @@ public extension ResultType {
 	}
 }
 
+/// Protocol used to constrain a version of `map` for `transform`s that throw.
+public protocol ErrorTypeConvertible: ErrorType {
+	typealias ConvertibleType = Self
+	static func errorFromErrorType(error:ErrorType) -> ConvertibleType
+}
+
+public extension ResultType where Error: ErrorTypeConvertible {
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
+	public func map<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+		return flatMap { value in
+			do {
+				return .Success(try transform(value))
+			} catch {
+				return .Failure(Error.errorFromErrorType(error) as! Error)
+			}
+		}
+	}
+}
+
 // MARK: - Operators
 
 infix operator &&& {
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -79,9 +79,8 @@ public extension ResultType where Error: ErrorTypeConvertible {
 				return .Success(try transform(value))
 			}
 			catch {
-				guard let convertedError = Error.errorFromErrorType(error) as? Error else {
-					fatalError("Unable to convert error type '\(error.dynamicType)' to type '\(Error.self)'")
-				}
+				let convertedError = Error.errorFromErrorType(error) as! Error
+				// Revisit this in a future version of Swift. https://twitter.com/jckarter/status/672931114944696321
 				return .Failure(convertedError)
 			}
 		}
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -67,7 +67,7 @@ public extension ResultType {
 /// Protocol used to constrain a version of `map` for `transform`s that throw.
 public protocol ErrorTypeConvertible: ErrorType {
 	typealias ConvertibleType = Self
-	static func errorFromErrorType(error:ErrorType) -> ConvertibleType
+	static func errorFromErrorType(error: ErrorType) -> ConvertibleType
 }
 
 public extension ResultType where Error: ErrorTypeConvertible {
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -77,8 +77,12 @@ public extension ResultType where Error: ErrorTypeConvertible {
 		return flatMap { value in
 			do {
 				return .Success(try transform(value))
-			} catch {
-				return .Failure(Error.errorFromErrorType(error) as! Error)
+			}
+			catch {
+				guard let convertedError = Error.errorFromErrorType(error) as? Error else {
+					fatalError("Unable to convert error type '\(error.dynamicType)' to type '\(Error.self)'")
+				}
+				return .Failure(convertedError)
 			}
 		}
 	}
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -64,7 +64,7 @@ public extension ResultType {
 	}
 }
 
-/// Protocol used to constrain a version of `map` for `transform`s that throw.
+/// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
 public protocol ErrorTypeConvertible: ErrorType {
 	typealias ConvertibleType = Self
 	static func errorFromErrorType(error: ErrorType) -> ConvertibleType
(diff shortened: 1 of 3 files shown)


Result/ResultType.swift
```diff
@@ -64,6 +64,26 @@ public extension ResultType {
...
+
+public extension ResultType where Error: ErrorTypeConvertible {
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
+	public func map<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+		return flatMap { value in
+			do {
+				return .Success(try transform(value))
+			} catch {
+				return .Failure(Error.errorFromErrorType(error) as! Error)
```
NachoSoto: This `as! Error` is scary.
Can you do instead
```swift
do {
  ...
} catch error as Error {
  return .Failure(Error.errorFromErrorType(error)
} catch error {
  // what do we do here? `fatalError` I guess? (which would be equivalent to this implementation)
}
```

bwhiteley: Good suggestion. `fatalError` would be better.

Result/ResultType.swift
```diff
@@ -64,6 +64,30 @@ public extension ResultType {
...
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
+	public func map<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+		return flatMap { value in
+			do {
+				return .Success(try transform(value))
+			}
+			catch {
+				guard let convertedError = Error.errorFromErrorType(error) as? Error else {
+					fatalError("Unable to convert error type '\(error.dynamicType)' to type '\(Error.self)'")
```
NachoSoto: Cool

bwhiteley:
What's the verdict on this PR?

Result/ResultType.swift
```diff
@@ -64,6 +64,30 @@ public extension ResultType {
...
+				}
+				return .Failure(convertedError)
```
robrix: Since this actually behaves like `flatMap` (but with a restricted parameter type), calling it `map` seems a bit confusing :confused:

bwhiteley: I agree, and would have preferred to call it `flatMap`. However, that resulted in "ambiguous use..." errors. Do you have a recommendation?

robrix: /cc @antitypical/result

gfontenot: Spitballing, but: Is there maybe a way to rewrite the current `flatMap` with `rethrows` in a way that covers both cases?

Result/ResultType.swift
```diff
@@ -64,6 +64,30 @@ public extension ResultType {
 	}
 }
 
+/// Protocol used to constrain a version of `map` for `transform`s that throw.
+public protocol ErrorTypeConvertible: ErrorType {
+	typealias ConvertibleType = Self
+	static func errorFromErrorType(error:ErrorType) -> ConvertibleType
```
NachoSoto: Nit: space between `error:` and `ErrorType`.

Result/ResultType.swift
```diff
@@ -64,6 +64,30 @@ public extension ResultType {
 	}
 }
 
+/// Protocol used to constrain a version of `map` for `transform`s that throw.
+public protocol ErrorTypeConvertible: ErrorType {
+	typealias ConvertibleType = Self
```
NachoSoto: Is this necessary? Maybe just use `Self`?

Result/ResultType.swift
bwhiteley: I don't think so. The return type of the `transform` is quite different, and part of the reason for this PR is to be able to pass in functions that are not aware of `Result`.

Result/ResultType.swift
bwhiteley: It is necessary because NSError is not final. From what I've seen, this is a common pattern.

Result/ResultType.swift
robrix: Yeah, I don’t think this is precisely `map` _or_ `flatMap`.

bwhiteley: I originally called it `tryMap` (PR #114), but tried to roll it into `flatMap` or `map` based on feedback in that PR. How are we feeling about `tryMap`?

neilpa: :+1: on `tryMap`

robrix: Yeah, that was my bad, my apologies. I’m :+1: on the name `tryMap`.

bwhiteley: I renamed it to `tryMap`. Any other concerns?

robrix:
:+1: from me. @NachoSoto?

Result/ResultType.swift
```diff
@@ -64,6 +64,30 @@ public extension ResultType {
...
+public extension ResultType where Error: ErrorTypeConvertible {
+
+	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
+	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
+		return flatMap { value in
+			do {
+				return .Success(try transform(value))
+			}
+			catch {
+				guard let convertedError = Error.errorFromErrorType(error) as? Error else {
```
NachoSoto: I'm confused about this, if `Error` conforms to `ErrorTypeConvertible`, this is always going to be `non-nil` **and** of type `Error`, right?

bwhiteley: You would think so, but 
```
ResultType.swift:82:58: error: 'Self.Error.ConvertibleType' is not convertible to 'Self.Error'; did you mean to use 'as!' to force downcast?
                                let convertedError = Error.errorFromErrorType(error) as Error
                                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
                                                                                     as!
```

NachoSoto: That seems like a bug.
What if you remove `ConvertibleType` and change `ErrorTypeConvertible`:
```swift
public protocol ErrorTypeConvertible: ErrorType {
    static func errorFromErrorType(error: ErrorType) -> Self
}
```

bwhiteley: That does permit the use of a plain `as` in `tryMap`, but then I can't make `NSError` conform to `ErrorTypeConvertible`, presumably because it's not `final`. 

I boiled it down to a sample that you can tinker with in a Playground. 

```
import Foundation

// For types that can be created from an arbitrary ErrorType
public protocol ErrorTypeConvertible: ErrorType {
    static func errorFromErrorType(error: ErrorType) -> ConvertibleType
}

protocol MyType {
    typealias Error: ErrorType
    var error: Error { get }
}

extension MyType where Error: ErrorTypeConvertible {
    func foo() -> Error {
        return Error.errorFromErrorType(error) as! Error // ** Why does this require `!`? **
    }
}

extension NSError: ErrorTypeConvertible {
    public static func errorFromErrorType(error: ErrorType) -> NSError {
        return error as NSError
    }
}
```

NachoSoto: Probably related to https://twitter.com/jckarter/status/672931114944696321.

Given that the cast can never fail I'd just use `as!` and add a comment to improve this later on.

NachoSoto:
:+1: thanks @bwhiteley! this looks good now :) :shipit:

PR #120, opened 2015-11-15
add a non-closure version of init

### LINKED ISSUE #119: Cannot use auto closure init in a closure
for example:
```swift
func(){
let result: Result<String, NSError> = Result(try tryIsSuccess("success")) //This is OK in function
}

let simpleBlock: ()->Void =  {
let result: Result<String, NSError> = Result(try tryIsSuccess("success")) //This is not OK in colure
}
simpleBlock()
```

So my suggestion is to add a init version without @autoclosure mark
```swift
extension Result{
    public init(_ f: () throws -> T) {
        self.init(try f())
    }
}
```

frogcjn:
add a non-closure version of init

Fixes #119

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -31,6 +31,15 @@ public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, Cu
 		}
 	}
 	
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws, non-autoclosure version
+    	public init?(_ f: () throws -> T?) {
+        	do {
+            		guard let result = try f() else { return nil }
+            		self = .Success(result)
+		} catch {
+        		self = .Failure(error as! Error)
+        	}
+        }
 
 	// MARK: Deconstruction
 


Result/Result.swift
```diff
@@ -31,6 +31,15 @@ public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, Cu
 		}
 	}
 	
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws, non-autoclosure version
+    	public init?(_ f: () throws -> T?) {
+        	do {
+            		guard let result = try f() else { return nil }
```
NachoSoto: I'm confused I think this should be defined in terms of `T`, not `T?`?

```swift
    	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws, non-autoclosure version
    	public init(_ f: () throws -> T) {
        	do {
            		self = .Success(try f())
		} catch {
        		self = .Failure(error as! Error)
        	}
        }
```

Then I think the `@autoclosure` version can be reimplemented based on this one :)

frogcjn: You can see https://github.com/antitypical/Result/issues/119#issuecomment-156814811

There is a case that I don't want to make a result.

NachoSoto: But that's a different constructor then. I thought this was to work around the fact that you can't use the `@autoclosure` method with a closure.

frogcjn: Yes, I know. Even if there is no bag in Swift compiler, there is still a need of un-autoclosure version.

frogcjn: Yes, I know. Even if there was no bag in Swift compiler, there is still a need of un-autoclosure version.

gfontenot: I agree with @NachoSoto. This should be defined in terms of `T` (and probably marked as `@noescape`). See the [materialize implementations](https://github.com/frogcjn/Result/blob/patch-1/Result/Result.swift#L158-L168) for reference.

frogcjn:
There is a use case in my code. I have provided it.

Result/Result.swift
neilpa: Why not make a `Break.Cancelled` error, return that and filter it out? I don't think it makes sense to add a failable initializer to `Result`.

gfontenot:
I don't think the example you gave is a compelling reason to add a failable initializer to `Result`, to be honest. Judging solely from the snippet, it seems like it should be possible to avoid `throws` entirely and construct the `Result` manually.

My understanding of the intent behind the functions converting `throws` to `Result` is that they are intended to be used with methods and functions that are out of your control, as a way to coerce those functions back into the world of `Result`. I don't think we intend to promote `throws` as a common control flow for constructing `Result` instances.

That being said, I _do_ see the need for the non-autoclosure version of the initializer, and would also be happy to help you figure out a way to write this without needing to rely on `throws`. For this PR, though, I think we should stick to creating a version of the constructor that doesn't use `@autoclosure` and instead uses `@noescape`. I also think we should convert the existing `@autoclosure` initializer to be written in terms of this one.

robrix:
Non-`@autoclosure` constructor added in #125.

PR #122, opened 2015-11-27
Add 'Reference Status' from versioneye to README

phimage:
Result is µframework and I expect that more people will use it instead of reinvent the wheel (like I see in some repository)
Versioneye give you the number of project that use your project as dependency like ReactiveCocoa, BrightFutures, ...
https://www.versioneye.com/objective-c/result/references
https://www.versioneye.com/objective-c/result/
badge


robrix:
Interesting! I’m :+1:, although I wonder if this # will be misleading due to the transitive inclusions from users of RAC &c.

Either way, thanks for the PR; I’m going to merge it.

PR #125, opened 2015-12-01
Added non-@autoclosure version of init.

### LINKED ISSUE #119: Cannot use auto closure init in a closure

NachoSoto:
Fixes #119. Alternative approach to #120.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -17,12 +17,12 @@ public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, Cu
 		self = .Failure(error)
 	}
 
-	/// Constructs a result from an Optional, failing with `Error` if `nil`
+	/// Constructs a result from an Optional, failing with `Error` if `nil`.
 	public init(_ value: T?, @autoclosure failWith: () -> Error) {
 		self = value.map(Result.Success) ?? .Failure(failWith())
 	}
 
-	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
 	public init(@autoclosure _ f: () throws -> T) {
 		self.init(f: f)
 	}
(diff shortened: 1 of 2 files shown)


Result/Result.swift
```diff
@@ -17,24 +17,28 @@ public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, Cu
...
 	}
 
-	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
 	public init(@autoclosure _ f: () throws -> T) {
+		self.init(f: f)
+	}
+
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
+	public init(@noescape f: () throws -> T) {
```
NachoSoto: I don't think we can make this a pure overload (as in, without the named parameter) and make the other constructor unambiguously delegate to this. It's actually quite disappointing that Swift can't even infer at compile time that doing `self.init(f)` is an infinite loop.

gfontenot:
:+1: from me

Result/Result.swift
robrix: Can we give the parameter a better name than `f` at least?

Result/Result.swift
```diff
@@ -17,24 +17,28 @@ public enum Result<T, Error: ErrorType>: ResultType, CustomStringConvertible, Cu
...
 	}
 
-	/// Constructs a result from an Optional, failing with `Error` if `nil`
+	/// Constructs a result from an Optional, failing with `Error` if `nil`.
 	public init(_ value: T?, @autoclosure failWith: () -> Error) {
 		self = value.map(Result.Success) ?? .Failure(failWith())
 	}
 
-	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws
+	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
```
robrix: Thank you. These have been bothering me :grin:

Result/Result.swift
NachoSoto: Sure thing. Other ideas:
- block
- function
- closure
- attempt

robrix: I think `attempt` sounds great :+1:

NachoSoto: Done

robrix:
:heart: this, @NachoSoto.

PR #129, opened 2015-12-10
Update version in podspec to 1.0.1

Thomvis:
We forgot to do this for the 1.0.1 release. The change is quite straight forward, but to make sure I'm not forgetting anything (like last time), I'd like a second pair of :eyes: to :+1:. Thanks!

### DIFF
--- a/Result.podspec
+++ b/Result.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'Result'
-  s.version      = '1.0.0'
+  s.version      = '1.0.1'
   s.summary      = 'Swift type modelling the success/failure of arbitrary operations'
 
   s.homepage     = 'https://github.com/antitypical/Result'


kylef:
Since there are new features, shouldn't it actually become 1.1.0 as per [semantic versioning](http://semver.org)?

> Given a version number MAJOR.MINOR.PATCH, increment the:

> MAJOR version when you make incompatible API changes,
> MINOR version when you add functionality in a backwards-compatible manner, and
> PATCH version when you make backwards-compatible bug fixes.

robrix:
Either way we should make sure the podspec & other notions of version (tags, etc) are in sync.

gfontenot:
> shouldn't it actually become 1.1.0 as per semantic versioning?

It should have been 1.1.0, yeah. This is my fault. I suggested 1.0.1 because I didn't realize there were new features.

That being said, I agree that we should stay in sync with the tagged version. We'll try to do better with versioning moving forward.

Thomvis:
So we should merge this in to make the 1.0.1 release complete, right?

NachoSoto:
That's my bad, I should have realized when I made the release but I was on autopilot :/

robrix:
:+1:, merging as-is.

gfontenot:
Pushed to trunk

PR #132, opened 2015-12-15
Added NoError type

### LINKED ISSUE #130: RFC: Add NoError type
```swift
enum NoError: ErrorType { }
```
As mentioned in [`Commandant`](https://github.com/Carthage/Commandant/pull/41/files#r47448896), I think it would be a good idea to put this type in `Result` so that `ReactiveCocoa`, `Commandant`, and other frameworks can take advantage of it.

NachoSoto:
Fixes #130

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -214,5 +214,13 @@ extension NSError: ErrorTypeConvertible {
 	}
 }
 
+// MARK: -
+
+/// An “error” that is impossible to construct.
+///
+/// This can be used to describe `Result`s where failures will never
+/// be generated. For example, `Result<Int, NoError>` describes a result that
+/// contains an `Int`eger and is guaranteed never to be a `Failure`.
+public enum NoError: ErrorType { }
 
 import Foundation


gfontenot:
Is it weird to hide this away at the bottom of `Result.swift`?

This seems to be a general pattern we're using so it's probably not worth holding up this PR for. I do wonder if we shouldn't do a pass at better file organization though.

gfontenot:
Anyway, I dig the change. Good stuff.

JaviSoto:
Is it worth exposing an extension on `Result where Error == NoError` with a var that returns a non-optional T value?

Other than that :+1: this is great!

NachoSoto:
> Is it weird to hide this away at the bottom of Result.swift?

Good point, I didn't consider that. It's definitely more important than that! I'd consider how it *looks* when looking at the generated "header" Swift module though.

NachoSoto:
> Is it worth exposing an extension on Result where Error == NoError with a var that returns a non-optional T value?

Ooooh good idea!

robrix:
Lovely :heart:

NachoSoto:
What do people think of the 2 questions?
- Location: @gfontenot where would you recommend placing it instead?
- non-optional `value` in an extension with `where Error == NoError`?

gfontenot:
I don't know that I have a better recommendation right now (other than maybe just a `NoError.swift` file). I'm OK with not changing anything here and instead doing a larger audit on the organization later.

gfontenot:
Oh, and I love the idea of a non-optional `value` property. Should it be named something else to avoid ambiguity?

NachoSoto:
It can probably share the same name (because of lvalue type inference), but it would mean that some users would have to explicitly specify the type.

I filed [this radar](http://www.openradar.me/23432210) (dupes appreciated! maybe I should submit to `swift-evaluation`?) precisely for this reason. Ideally the optional type would not exist for `NoError`.

mdiep:
I'm going to go ahead and merge this as-is. I don't think the location is terribly important (and can be changed later) and Swift doesn't make a non-optional variant very easy.

NachoSoto:
:heart_eyes:

Issue #124, opened 2015-11-29
Conveniences ought to have a home


robrix:
We’ve had a few PRs along the lines of #123, where someone has submitted a scratch for a not-uncommon itch. We’ve been turning them down—mostly—thus far, but maybe we should instead be finding them a better home.

As I understand it, this is the sort of thing that `git`’s [`contrib` dir](https://git.kernel.org/cgit/git/git.git/tree/contrib/README) exists to enable. Maybe we should consider something similar. Some ideas:

- a `Contrib` (sub?)framework encapsulating conveniences and experiments
- a section of the readme linking to gists with these things
- just including them in the framework and being done with it

NachoSoto:
I like the idea! Especially as a separate framework.

gfontenot:
I love the idea of linking to gists or providing a `Contrib` framework alongside `Result`. I don't think we should be including them in the main framework though.

neilpa:
The one problem with a `Contrib` style framework in the same repo is it doesn't play all that nice with `carthage`.

NachoSoto:
I was thinking it'd be a different repo for that reason.

Thomvis:
Agreed, a separate framework & repo seems the best approach. In my experience however it is easy for the two frameworks to get out of sync, e.g. If you forget to update the contrib after a PR in the main repo. Is there a proven process for this?

Just my 5 cents for the name: `ResultOriented.framework`

NachoSoto:
> If you forget to update the contrib after a PR in the main repo. Is there a proven process for this?

Well the "contrib" version would reference a specific `Result` version so compatibility would always be explicit.

That also means that one can simply change their `Carthage` reference from `Result` to this other framework, and their existing usages of `Result` would continue to work.

Thomvis:
> Well the "contrib" version would reference a specific Result version so compatibility would always be explicit.

Understood. I was hoping there was some sort of safeguard/check that would prevent the contrib library from lagging behind.

robrix:
@Thomvis: We could add it as a private dependency and then test it.

kylef:
> In my experience however it is easy for the two frameworks to get out of sync, e.g. If you forget to update the contrib after a PR in the main repo.

Since Result just hit 1.0, I'd expect the API to be pretty stable and at least no backwards incompatible changes anytime soon. So I wouldn't expect this to be a problem.

robrix:
Can we provide the contrib stuff in a submodule of the Result module? e.g.:

```swift
[...]

PR #133, opened 2015-12-28
Implemented following two array helper functions with corresponding unit tests

chainone:
MapFilter: Return a array of `V` by applying transform to all elements
of array 'U' and filtering out failure ones

MapM: Return a Result with an array of `V`s if all Results of applying
`transform` are `Success`es or return the error of the first `Failure`

### DIFF
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -101,3 +101,29 @@ infix operator &&& {
 public func &&& <L: ResultType, R: ResultType where L.Error == R.Error> (left: L, @autoclosure right: () -> R) -> Result<(L.Value, R.Value), L.Error> {
 	return left.flatMap { left in right().map { right in (left, right) } }
 }
+
+/// Return a array of `V` by applying transform to all elements of array 'U' and filtering out failure ones
+public func mapFilter<U, V, Error>(results: [U], @noescape transform: U -> Result<V, Error>) -> [V]{
+	var vs = [V]()
+	for result in results{
+		if case let .Success(value) = transform(result){
+			vs.append(value)
+		}
+	}
+	return vs
+}
+
+/// Return a Result with an array of `V`s if all Results of applying `transform` are `Success`es or return the error of the first `Failure`
+public func mapM<U, V, Error>(results: [U], @noescape transform: U -> Result<V, Error>) -> Result<[V], Error>{
+	var vs = [V]()
+	for result in results{
+		let mapped = transform(result)
+		switch mapped{
+		case let .Success(value):
+			vs.append(value)
+		case let .Failure(error):
+			return Result(error: error)
+		}
+	}
+	return Result(vs)
+}
--- a/Tests/ResultTests.swift
+++ b/Tests/ResultTests.swift
@@ -132,6 +132,28 @@ final class ResultTests: XCTestCase {
 		let resultFailureRight = success &&& failure2
 		XCTAssert(resultFailureRight.error == error2)
 	}
+	
+	func testMapFilterOnArray(){
+		let a = [1,2,3,4,5,6,7,8,9]
+		let r = mapFilter(a, transform: evenClosure)
+
+		XCTAssert(r == ["2","4","6","8"])
+	}
+	
+	func testMapMOnArraySuccess(){
+		let a = [2,4,6,8]
+		if case let .Success(r) = mapM(a, transform: evenClosure){
+			XCTAssert(r == ["2","4","6","8"])
+		}else{
+			XCTFail()
+		}
+	}
+	
+	func testMapMOnArrayFailure(){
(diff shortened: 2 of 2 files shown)


Result/ResultType.swift
```diff
@@ -101,3 +101,29 @@ infix operator &&& {
...
+		let mapped = transform(result)
+		switch mapped{
+		case let .Success(value):
+			vs.append(value)
+		case let .Failure(error):
+			return Result(error: error)
+		}
+	}
+	return Result(vs)
+}
```
robrix: I think both of these would be more idiomatic if expressed as extension methods on `CollectionType`. Not only more idiomatic, but they’d also be available on collections other than `Array`.

Result/ResultType.swift
```diff
@@ -101,3 +101,29 @@ infix operator &&& {
 public func &&& <L: ResultType, R: ResultType where L.Error == R.Error> (left: L, @autoclosure right: () -> R) -> Result<(L.Value, R.Value), L.Error> {
 	return left.flatMap { left in right().map { right in (left, right) } }
 }
+
+/// Return a array of `V` by applying transform to all elements of array 'U' and filtering out failure ones
+public func mapFilter<U, V, Error>(results: [U], @noescape transform: U -> Result<V, Error>) -> [V]{
```
robrix: If expressed as an extension method on `CollectionType`, I think this becomes essentially equivalent to a straightforward use of the collection-to-optional variant of `flatMap` in the standard library:

```swift
values.mapFilter(transform)
// is equivalent to
values.flatMap { transform($0).value }
```

In which case maybe we’d be better to just use the standard library’s functionality and not extend it with this particular method.

Tests/ResultTests.swift
```diff
@@ -132,6 +132,28 @@ final class ResultTests: XCTestCase {
...
+		}else{
+			XCTFail()
+		}
+	}
+	
+	func testMapMOnArrayFailure(){
+		let a = [2,4,5,8]
+		let r = mapM(a, transform: evenClosure)
+		XCTAssert(r.error == error)
+	}
```
robrix: Thank you for tests! :bow:

Result/ResultType.swift
```diff
@@ -101,3 +101,29 @@ infix operator &&& {
...
+	for result in results{
+		if case let .Success(value) = transform(result){
+			vs.append(value)
+		}
+	}
+	return vs
+}
+
+/// Return a Result with an array of `V`s if all Results of applying `transform` are `Success`es or return the error of the first `Failure`
+public func mapM<U, V, Error>(results: [U], @noescape transform: U -> Result<V, Error>) -> Result<[V], Error>{
```
robrix: I’d prefer a name somewhat less obscure than `mapM` (tho I acknowledge its [`Traversable` heritage](http://hackage.haskell.org/package/base-4.8.1.0/docs/Data-Traversable.html#v:mapM)).

Perhaps `all` or `allOrNothing`?

robrix: A little further on this, and similar to [the discussion of `mapFilter` above](https://github.com/antitypical/Result/pull/133/files#r48482363), we can probably decouple the essential action here, that of producing a success with all values if they are all successful, or failing otherwise, from the incidental action of mapping using `transform`, to get a slightly more idiomatic take on it:

```swift
public extension CollectionType where Generator.Element: ResultType {
    public func all() -> Result<[Generator.Element.Value], Generator.Element.Error> {
        …
    }
}
```

`mapM`’s full behaviour can be recovered by chaining after `CollectionType.map`:

```swift
let results = values.map(validate).all()
```

And you can use `lazy`, as well.

robrix:
Thanks for the pull request! :bow:

This functionality is quite useful. I’ve left a few notes above, but you might not want to address any/all of them straight away since there’s some question about where this should end up living.

We’ve typically shied away from including extensions on other types in Result, but [we’ve been discussing](https://github.com/antitypical/Result/issues/124) providing a home for them, possibly in a separate framework.

@antitypical/result, what do you think? Is this a good motivator for a `contrib` framework or similar, or does this belong in Result proper?

NachoSoto:
I agree, this is useful but it seems out of the scope of `Result` itself.

Result/ResultType.swift
chainone: Great response, @robrix !
And I agree that a particular `mapFilter` is not necessary here.

BTW: my first version of `mapFilter' is something like 'values.lazy.map(transform).filter(...).map{$0!}' which is far more complicated than the version your provided.
:+1

Result/ResultType.swift
chainone: I agree that:
1. Providing a lower level "all()" is better than a higher level of "mapM" here
2. Due to the nature of "all()", lazy is preferred.

PR #138, opened 2016-02-24
Fix Xcode 7.3 Swift warnings

maxsz:


kylef:
Doesn't these changes make Result unusable on Swift 2.1 (current stable version)? Swift 3.0 won't be out for a while yet.

mdiep:
Yeah, these are breaking changes, so we'll want to hold off on them for now.

larryonoff:
just to check that It's time to merge these changes!

ishkawa:
I think we should provide Result without warning for Swift 2.2 developers, because there are several inconveniences caused by warnings. For example, `pod lib lint` treats warnings as errors, so libraries depending on Result cannot push its podspec to the trunk.

Here's an example of result of `pod spec lint`:
```
$ pod lib lint
 -> APIKit (2.0.0-beta.2)
    - WARN  | xcodebuild:  Result/Result/Result.swift:93:70: warning: __FUNCTION__ is deprecated and will be removed in Swift 3, please use #function
    - WARN  | xcodebuild:  Result/Result/Result.swift:93:99: warning: __FILE__ is deprecated and will be removed in Swift 3, please use #file
    - WARN  | xcodebuild:  Result/Result/Result.swift:93:121: warning: __LINE__ is deprecated and will be removed in Swift 3, please use #line
    - WARN  | xcodebuild:  Result/Result/Result.swift:172:41: warning: __FUNCTION__ is deprecated and will be removed in Swift 3, please use #function
    - WARN  | xcodebuild:  Result/Result/Result.swift:172:70: warning: __FILE__ is deprecated and will be removed in Swift 3, please use #file
    - WARN  | xcodebuild:  Result/Result/Result.swift:172:92: warning: __LINE__ is deprecated and will be removed in Swift 3, please use #line
    - WARN  | xcodebuild:  Result/Result/Result.swift:182:38: warning: __FUNCTION__ is deprecated and will be removed in Swift 3, please use #function
    - WARN  | xcodebuild:  Result/Result/Result.swift:182:67: warning: __FILE__ is deprecated and will be removed in Swift 3, please use #file
    - WARN  | xcodebuild:  Result/Result/Result.swift:182:89: warning: __LINE__ is deprecated and will be removed in Swift 3, please use #line
    - WARN  | xcodebuild:  Result/Result/ResultType.swift:5:2: warning: use of 'typealias' to declare associated types is deprecated; use 'associatedtype' instead
    - WARN  | xcodebuild:  Result/Result/ResultType.swift:6:2: warning: use of 'typealias' to declare associated types is deprecated; use 'associatedtype' instead
    - WARN  | xcodebuild:  Result/Result/ResultType.swift:69:2: warning: use of 'typealias' to declare associated types is deprecated; use 'associatedtype' instead
[!] APIKit did not pass validation, due to 12 warnings (but you can use `--allow-warnings` to ignore them).
```

Since it introduces breaking changes, the new version might be 2.0.

@antitypical/result What do you think?

robrix:
Agreed:

1. Let’s try to avoid warnings.
2. This is going to have to be 2.0.

mdiep:
As a counterpoint:

I'd argue that we should maintain compatibility with both Swift 2.1 and Swift 2.2 for a single release if possible. Upgrading Xcode versions is a significant headache; partly because you need to upgrade the whole world. Maintaining compatibility with both for a single release will ease the upgrade path for all consumers—even if we immediately released both versions.

But I'm also not sympathetic to CocoaPods or the problems it creates.

mdiep:
Actually, it looks like 1.0.2 builds with Xcode 7.3, so I'm content with Swift 2.2 compatibility only in the new release. :+1:

paulz:
Would it be possible to update travis to use Xcode 7.3 to get this merged into new release?
See our issue:  https://github.com/antitypical/Result/issues/142#issuecomment-200660384

robrix:
> I'd argue that we should maintain compatibility with both Swift 2.1 and Swift 2.2 for a single release if possible. Upgrading Xcode versions is a significant headache; partly because you need to upgrade the whole world. Maintaining compatibility with both for a single release will ease the upgrade path for all consumers—even if we immediately released both versions.

@mdiep: That’s a good point. Should we address that in #141 before it :ship:s, then?

mdiep:
The current release (1.0.2) builds with both Swift 2.1 and 2.2, so we're all set. No work necessary.

robrix:
:+1: ok, thanks.

mdiep:
All the changes here were already merged to master in #141.

PR #140, opened 2016-03-07
Run `swift test` on Linux

norio-nomura:
- Use `#if` only on codes that can't be compiled.
- Add printing message to crashing test on Linux and run them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
       language: objective-c
     - script:
         - swift build
+        - swift test
       env: JOB=Linux
       sudo: required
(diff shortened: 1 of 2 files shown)


robrix:
Nice! Thank you :grin:

robrix:
@norio-nomura: I’ve invited you to the repo. We use a PR-based workflow for the core contributors just like you’ve been doing already, but this way it should be slightly more convenient for you to e.g. tag a release if/when you need to.

Thanks again for your contributions!

norio-nomura:
Thanks for merging and invitation! :pray:

PR #141, opened 2016-03-23
Swift 2.2 and 3.0 compatibility

dileping:


Result/Result.swift
```diff
@@ -217,13 +217,19 @@ public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T
...
-	public static func errorFromErrorType(error: ErrorType) -> NSError {
-		return error as NSError
+	
+	public extension ErrorTypeConvertible where Self : NSError {
+		public func force<T>() -> T {
+			return self as! T
+		}
+	}
+	
+	extension NSError : ErrorTypeConvertible {
```
NachoSoto: There should be no space after `NSError`.

Result/ResultType.swift
```diff
@@ -1,9 +1,20 @@
...
 
+#if swift(>=3.0)
+	public typealias ResultErrorType = ErrorProtocol
+#else
+	public typealias ResultErrorType = ErrorType
+#endif
+
 /// A type that can represent either failure with an error or success with a result value.
 public protocol ResultType {
+	#if swift(>=2.2)
```
NachoSoto: This can't be parsed prior to Swift 2.2, so it's useless to try to maintain compatibility with previous versions. Like I mentioned in https://github.com/ReactiveCocoa/ReactiveCocoa/pull/2760/files/6f63eb6569772a90f0b4358acb33b93a4f8105b2#discussion_r56916728, it might not be worth the headache of trying to maintain backwards compatibility. There's already a stable version released for Swift 2.1.x, I see no reason to make the codebase support old versions forever.

Tests/Result/ResultTests.swift
```diff
@@ -75,7 +75,7 @@ final class ResultTests: XCTestCase {
 		let result1 = materialize(try tryIsSuccess("success"))
 		XCTAssert(result1 == success)
 
-		let result2 = materialize { try tryIsSuccess("success") }
+		let result2:Result<String, NSError> = materialize { try tryIsSuccess("success") }
```
NachoSoto: Space after `:`

Tests/Result/ResultTests.swift
```diff
@@ -87,7 +87,7 @@ final class ResultTests: XCTestCase {
 			let result1 = materialize(try tryIsSuccess(nil))
 			XCTAssert(result1.error == error)
 
-			let result2 = materialize { try tryIsSuccess(nil) }
+			let result2:Result<String, NSError> = materialize { try tryIsSuccess(nil) }
```
NachoSoto: Ditto

dileping:
@NachoSoto fixed your comments

robrix:
Does this overlap with/replace #138?

Result/ResultType.swift
```diff
@@ -65,9 +71,9 @@ public extension ResultType {
 }
 
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
-public protocol ErrorTypeConvertible: ErrorType {
+public protocol ErrorTypeConvertible: ResultErrorType {
 	typealias ConvertibleType = Self
-	static func errorFromErrorType(error: ErrorType) -> ConvertibleType
+	static func errorFromErrorType(error: ResultErrorType) -> Self
```
ikesyo: Why the return type is replaced with `Self`?

dileping:
It might overlap with #138 in one place, but in general it solves different issue. #138 fixes 2.2 warnings, but this one solves Swift 3.0 compatibility (read linux support).

Result/ResultType.swift
dileping: @ikesyo  Because it won't build otherwise with Swift 3.0

dileping:
In case this one will be merged I could work on providing warning-free version (same as #138) currently does

Result/ResultType.swift
ikesyo: What is the build failure? And if that is the case, `ConvertibleType` should not be needed anymore.

dileping: @ikesyo It looks like it is. The thing is that you have to explicitly put there `Self` in implementation. Try to install latest 3.0 toolchain and check. In case I'm wrong, I would be happy to remove it.

robrix:
> It might overlap with #138 in one place, but in general it solves different issue. #138 fixes 2.2 warnings, but this one solves Swift 3.0 compatibility (read linux support).

Thank you for clarifying :bow:

Result/Result.swift
```diff
@@ -217,13 +217,19 @@ public func >>- <T, U, Error> (result: Result<T, Error>, @noescape transform: T
 // MARK: - ErrorTypeConvertible conformance
 
 #if !os(Linux)
-
-/// Make NSError conform to ErrorTypeConvertible
-extension NSError: ErrorTypeConvertible {
-	public static func errorFromErrorType(error: ErrorType) -> NSError {
-		return error as NSError
+	
+	public extension ErrorTypeConvertible where Self : NSError {
```
NachoSoto: Indentation has changed here.

norio-nomura:
Should we support Swift 2.2 and 3 simultaneously in single branch?

NachoSoto:
> Should we support Swift 2.2 and 3 simultaneously in single branch?

THAT would be nice.

dileping:
So, guys. Do you think I need to do something in addition here or it is good enough as it is?

Result/Result.swift
dileping: @NachoSoto Don't you think this is a "more correct" indentation considering it's wrapped with `#if`?

dileping:
@norio-nomura 

> Should we support Swift 2.2 and 3 simultaneously in single branch?

Taking into consideration that the only usable option for Linux is Swift 3.0, I think it's great. We are using Result in our server-side software [Express](https://github.com/crossroadlabs/Express) and now since latest Linux snapshot release we are porting everything to support 2.2 and 3.0 simultaneously. Just for your consideration - Result is one of the easiest parts :)

norio-nomura:
> We are using Result in our server-side software Express and now since latest Linux snapshot release we are porting everything to support 2.2 and 3.0 simultaneously.

Oh, I understand. I have considered same approach on another repository and hesitate to do, because that means expecting dependencies to support 2.2 and 3.0 simultaneously.

> Do you think I need to do something in addition here or it is good enough as it is?

Test on Travis-CI is failing.
https://travis-ci.org/antitypical/Result/builds/117990150

Maybe we should test on 
- Xcode 7.3 with default toolchain
- Xcode 7.3 with `swift-DEVELOPMENT-SNAPSHOT-2016-03-16-a`
- linux with `swift-DEVELOPMENT-SNAPSHOT-2016-03-16-a`

`swift-DEVELOPMENT-SNAPSHOT-2016-03-16-a` needs workaround on [SR-975](https://bugs.swift.org/browse/SR-975)

dileping:
@norio-nomura OK, will handle it and get back to you.

Result/Result.swift
robrix: While I personally don’t care, I would note that sticking with the existing indentation is generally good practice.

robrix: Upon reflection, however, I suggest we leave this alone so as to expedite getting this PR merged.

dileping:
@norio-nomura Btw, if you are having the same issue with simultaneous support of 2.2 and 3.0 (and most probably Linux), here is something I'm working on currently: [Boilerplate](https://github.com/crossroadlabs/Boilerplate). The main idea is to keep there all the Swift incompatibilities and "insufficiency" in one single place.

It currently has workarounds for:
* Linux/OS X compatibility
* Bridging issues
* Some absent utility classes
* Swift 3.0 shims (so you can write in Swift 2.2 with the very same API as in Swift 3.0)

Current work in progress is for Swift 3.0 shims (contributions are very welcome as I currently add stuff there just when I encounter a problem in other projects). So it's kind of case-by-case, very practice oriented framework.

The only dependency is Result.

Have not done any README yet, but will be there in the best kind of Crossroad Labs style soon.

Hope it can help you with your stuff too.

dileping:
OK, guys. The warnings are fixed and tests are passing on both Mac and Linux. I have not added explicit Mac tests with 3.0 toolchain, because anyways it's tested as Linux version, but will make `travis.yml` much more cumbersome.

I think it's ready for merge unless you have any other suggestions or comments.

robrix:
This looks good to me; @NachoSoto?

NachoSoto:
Ship it! :star:

Result/ResultType.swift
```diff
@@ -65,9 +71,9 @@ public extension ResultType {
 }
 
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
-public protocol ErrorTypeConvertible: ErrorType {
-	typealias ConvertibleType = Self
-	static func errorFromErrorType(error: ErrorType) -> ConvertibleType
+public protocol ErrorTypeConvertible: ResultErrorType {
+	associatedtype ConvertibleType = Self
```
ikesyo: Why this still exists even though it is not used anywhere?

robrix: Ah, good catch. Could you please remove this before we merge, @dileping?

dileping: @ikesyo @robrix , I have just checked if we can get rid of it. Seems not, so leaving it as is.

robrix: :+1:

robrix:
Thank you @dileping!

dileping:
@robrix Welcome! Hope to see new tag soon ;)

PR #143, opened 2016-03-24
Add support for Swift 2.2 and Xcode 7.3

genericspecific:


robrix:
Thanks for the PR! I think we have this adequately covered by #141, however.

PR #144, opened 2016-03-25
Update README.md, remove inaccurate note

sprynmr:
Looks to me like materialize is available again, unless I'm missing something.


robrix:
Good catch.

PR #147, opened 2016-03-28
Reconsider ErrorTypeConvertible

ikesyo:
This was changed in https://github.com/antitypical/Result/pull/141. But the current form does not looks correct since the `ConvertibleType` associated type is not used in any method signature.

This change ~~brings back the original definition~~ removes `ConvertibleType` from `ErrorTypeConvertible` and worked well for both Swift 2.2 on OS X and swift-DEVELOPMENT-SNAPSHOT-2016-03-24-a on Linux.

### DIFF
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -72,11 +72,10 @@ public extension ResultType {
 
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
 public protocol ErrorTypeConvertible: ResultErrorType {
-	associatedtype ConvertibleType = Self
-	static func errorFromErrorType(error: ResultErrorType) -> ConvertibleType
+	static func errorFromErrorType(error: ResultErrorType) -> Self
 }
 
-public extension ResultType where Error: ErrorTypeConvertible, Error.ConvertibleType == Error {
+public extension ResultType where Error: ErrorTypeConvertible {
 
 	/// Returns the result of applying `transform` to `Success`es’ values, or wrapping thrown errors.
 	public func tryMap<U>(@noescape transform: Value throws -> U) -> Result<U, Error> {
--- a/Result/ResultType.swift
+++ b/Result/ResultType.swift
@@ -73,10 +73,10 @@ public extension ResultType {
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
 public protocol ErrorTypeConvertible: ResultErrorType {
 	associatedtype ConvertibleType = Self
-	static func errorFromErrorType(error: ResultErrorType) -> Self
+	static func errorFromErrorType(error: ResultErrorType) -> ConvertibleType
 }
 
-public extension ResultType where Error: ErrorTypeConvertible {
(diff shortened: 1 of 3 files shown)


Result/ResultType.swift
```diff
@@ -73,10 +73,10 @@ public extension ResultType {
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
 public protocol ErrorTypeConvertible: ResultErrorType {
 	associatedtype ConvertibleType = Self
-	static func errorFromErrorType(error: ResultErrorType) -> Self
+	static func errorFromErrorType(error: ResultErrorType) -> ConvertibleType
 }
 
-public extension ResultType where Error: ErrorTypeConvertible {
+public extension ResultType where Error: ErrorTypeConvertible, Error.ConvertibleType == Error {
```
ikesyo: I added the constraint to make [the implementation here](https://github.com/antitypical/Result/pull/147/files#diff-77b13e3ea6c630c28c5afd2ccbd7904dR88) safer.

ikesyo:
This is breaking API changes, but Result 2.0 is already released. So should this be shipped as 3.0?

norio-nomura:
I did not use pre-release identifier of semver on 2.0.0 because of SwiftPM's current behavior.
But, I should mark 2.0.0 as Pre-release on https://github.com/antitypical/Result/releases/tag/2.0.0.
May I change that now?

ikesyo:
Just setting pre-release checkbox on at the GitHub release would not have any effects for SwiftPM or Carthage or CocoaPods, I think.

norio-nomura:
If we would mark 2.0.0 as pre-release, should we continue new releases as pre-release until the release of Swift 3?
I think we may make breaking changes until the release of Swift 3. If we would not mark them as pre-release, we may need to incrementing major version on them.

ikesyo:
But unfortunately, [2.0.0 release](https://github.com/antitypical/Result/releases/tag/2.0.0) **is already shipped as non-pre-release version**, because the release has no pre-release identifiers as you said. So just adding pre-release mark on GitHub would not help anything.

norio-nomura:
I'm so sorry that I did not tag 2.0.0 as pre-release.

norio-nomura:
Is `associatedtype ConvertibleType` still needed?

norio-nomura:
> Is `associatedtype ConvertibleType` still needed?

It seems that is still needed.

norio-nomura:
I have an interest in the codes that would be affected by this change.

ikesyo:
I revisited this, and came to the thought that removing `ConvertibleType` and `public func force<T>() -> T`. I believe the latter should not be exposed as a public API and the leakage should be a bug. Considering that, this change is just a bug fix and can be shipped as 2.0.1.

@antitypical/result Any thoughts?

robrix:
@ikesyo: Agreed :100:% re: `force`.

ikesyo:
Can I merge this?

robrix:
Looks good to me! :+1:

PR #149, opened 2016-04-13
Update to swift-DEVELOPMENT-SNAPSHOT-2016-04-12-a

norio-nomura:
Changes are affected by:
- [SE-0046 Establish consistent label behavior across all parameters including first labels](https://github.com/apple/swift-evolution/blob/master/proposals/0046-first-label.md)
- [SE-0055 Make unsafe pointer nullability explicit using Optional](https://github.com/apple/swift-evolution/blob/master/proposals/0055-optional-unsafe-pointers.md)


norio-nomura:
Should we continue to support Swift 2.2 and 3 simultaneously?

robrix:
> Should we continue to support Swift 2.2 and 3 simultaneously?

IMO yes, but I’m significantly less invested in this than I used to be so it’d be good to get others’ opinions. @antitypical/result: thoughts?

rnapier:
@robrix do you mean less invested in Swift 2/3, less invested in Swift, or less invested in Result? I've found myself mostly drifting away from Result in my own work. I just found the impedance mismatch with the rest of Cocoa too high to be worth the trouble most of the time. I still throw together a custom Result when I need one for some particular problem, but I almost always just code up the couple of pieces I need rather than using a library. Curious about your experience.

robrix:
I meant Swift; I don’t use it at all any more.

mdiep:
> Should we continue to support Swift 2.2 and 3 simultaneously?

I think so, if possible.

norio-nomura:
Thanks. I removed [WIP] tag.

robrix:
@norio-nomura: Looks good to me :+1:

mdiep:
Unfortunately, this needs a merge from master now.

norio-nomura:
Updated.

norio-nomura:
Thanks! :pray:

Issue #151, opened 2016-04-14
Add @warn_unused_result to APIs


mdiep:
/cc #150

robrix:
I feel like #152 covers this sufficiently.

Issue #150, opened 2016-04-14
Example of propagating errors


jozsef-vesza:
I've recently started experimenting with `Result`, and I would like to understand, how to propagate failures.

Consider the following method chain:

```swift
func importDataFromUrls(urls: [NSURL]) -> SignalProducer<MyType, MyErrorType> {
    return SignalProducer<MyType, MyErrorType> { (observer, disposable) in
        urls.forEach { (sourceURL) in
            self.createFolder(sourceURL)
                .flatMap { self.unarchiveContents($0, toURL: $1) }
                .flatMap { self.mapToModel($0) }
                .map { item in
                    observer.sendNext(item)
                }
        }
        
        observer.sendCompleted()
    }
}
```

The methods in the chain look like this:

```swift
func createFolder(source: NSURL) -> Result<FolderDescriptor, MyErrorType> {
    return fileManager.createDirectoryInApplicationSupport(source)
}

func unarchiveContents(sourceURL: NSURL, toURL: NSURL) -> Result<NSURL, MyErrorType> {
    return unarchiver.unzipFile(sourceURL, destination: toURL)
}

func mapToModel(contentURL: NSURL) -> Result<MyType, MyErrorType> {
    return fileManager.dataInDirectoryAtURL(contentURL).flatMap { descriptor in
        return Item.create(descriptor)
    }
}
```

I made a mock implementation for the first item of the chain, which should always fail:
```swift
func createFolder(source: NSURL) -> Result<DataDescriptor, MyErrorType> {
    return Result(error: .FileError)
}
```

If I understand `flatMap` correctly, successes will be transformed, and errors will be rewrapped. When I run my test, `unarchiveContents` won't get called, and the chain breaks. I either horribly misunderstand the whole chaining/rewrapping mechanism, or miss a step to propagate the error. 

Could anyone point me in the right direction?

mdiep:
The issue is that you're not doing anything with the result of your chain. You need to check the `Result`!

```swift
let result = self.createFolder(sourceURL)
                .flatMap { self.unarchiveContents($0, toURL: $1) }
                .flatMap { self.mapToModel($0) }
                .map { item in
                    observer.sendNext(item)
                }
// do something with result here
```

jozsef-vesza:
You're right, this one did the trick:

```swift
switch result {
case .Success(let item):
    observer.sendNext(item)
case .Failure(let error):
    observer.sendFailed(error)
}
```

Thank you!

PR #152, opened 2016-04-15
Warn when result is unused.

jozsef-vesza:
As a follow-up for #150, and to help with #151, I've annotated the API with `@warn_unused_result`. With this change the following line will trigger a warning:

```swift
doOperation().map { string in
    print(string)
}
```

If the result value is not used later on, it will also trigger a warning:
```swift
let result = doOperation().flatMap(doOtherOperation)
```


robrix:
Wonderful! Thank you @jozsef-vesza.

PR #153, opened 2016-04-15
Redefine functions in terms of ResultType

natestedman:
This redefines the functions `recover`, `recoverWith`, `==`, `!=`, `??`, and `>>-` in terms of `ResultType`, allowing them to be used more generically. The operators only used functionality that was already available in `ResultType`, so the only changes are to the type declarations for the functions - the implementations are identical.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -139,17 +139,6 @@ public enum Result<T, Error: ResultErrorType>: ResultType, CustomStringConvertib
 	}
 }
 
-
-/// Returns the value of `left` if it is a `Success`, or `right` otherwise. Short-circuits.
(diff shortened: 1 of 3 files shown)


mdiep:
Looks good to me, but needs a merge from master.

@antitypical/result Any objections to this?

robrix:
I think it’s a good idea; thanks @natestedman!

natestedman:
Updated against master - is the [SPM job](https://travis-ci.org/antitypical/Result/jobs/123715855) using Swift 3 code? Otherwise, need to verify those as I don't have Swift 3 installed locally.

mdiep:
The SPM and Linux jobs use the Swift development snapshot from the master branch, so they're as close to Swift 3 as you can get.

mdiep:
Thanks for the PR!

PR #163, opened 2016-06-14
Enabled Whole Module optimization

NachoSoto:
I've also removed all optimization settings in the targets, and just defined these once at the project level:

screen shot 2016-06-14 at 16 35 27

_Note: tests are still passing with this in Release mode_


mdiep:
✨

PR #164, opened 2016-06-14
Updated LastUpgradeCheck to Xcode 8

NachoSoto:
That way we're not asked to migrate again when opening the project, with suggestions like this:
screen shot 2016-06-14 at 16 29 01


ikesyo:
LGTM 👍 Thanks @NachoSoto! I forgot to do this in #162. 😅

Issue #166, opened 2016-06-26
Dropping Swift 2.x supports


norio-nomura:
I experienced difficulty with using master branch of antitypical/Result from Swift 3 migrated project.
Most of projects are using xcworkspace and Xcode 8 does not provide overriding configuration of referencing projects.
`SWIFT_VERSION = 2.3` is applied on project, but not workspace.

I think that we need swift-3 only branch of Result for resolving that.
So, I created swift-3 branch on my fork and dropped Swift 2.x supports on that.
https://github.com/norio-nomura/Result/tree/swift-3
I also changed some naming to conforming Swift 3.0 API guidelines on that.

@antitypical/result: How do you think about this?

mdiep:
Since Result has a release that supports Swift 2.3 / Xcode 8, I think we can adopt Swift 3 on `master` and start releasing alpha versions of Result 3.

NachoSoto:
I'm :+1: with that.

norio-nomura:
Thanks! I opened #167.

PR #167, opened 2016-06-26
Swift 3.0 Preview 1

norio-nomura:
#166
Summary:
- Drop Swift 2.x support
- Swift 3 API Design Guidelines
- Swift 3.0 Preview 1

API changes:
- `Result`
  - `.Success` to `.success`
  - `.Failure` to `.failure`
- `ResultType` to `ResultProtocol`
  - `recoverWith(_:)` to `recover(with:)`
- `ResultErrorType` to `ErrorProtocol`
- `ErrorTypeConvertible` to `ErrorProtocolConvertible`
  - `errorFromErrorType_:)` to `error(from:)`

Migration supporting codes are added to the end of each source files.


Result/ResultProtocol.swift
```diff
@@ -1,30 +1,20 @@
...
 
-#if swift(>=3.0)
-	public typealias ResultErrorType = ErrorProtocol
-#else
-	public typealias ResultErrorType = ErrorType
-#endif
-
 /// A type that can represent either failure with an error or success with a result value.
-public protocol ResultType {
+public protocol ResultProtocol {
```
NachoSoto: This change makes me sad

Result/ResultProtocol.swift
```diff
@@ -237,28 +145,40 @@ public func == <T: ResultType where T.Value: Equatable, T.Error: Equatable> (lef
...
-#if swift(>=3)
-public func ?? <T: ResultType> (left: T, right: @autoclosure () -> T) -> T {
-	return left.recoverWith(right())
+public func ?? <T: ResultProtocol> (left: T, right: @autoclosure () -> T) -> T {
+	return left.recover(with: right())
+}
+
+// MARK: - migration support
+@available(*, unavailable, renamed: "ResultProtocol")
+public typealias ResultType = ResultProtocol
```
NachoSoto: Awesome

NachoSoto:
:shipit:

Result/ResultProtocol.swift
robrix: For why?

NachoSoto: I think `Type` was a much better suffix, the fact that it's a protocols is already encoded in the type.

NachoSoto: (Not critiquing this change though, I know it's aligned with the new stdlib conventions.)

robrix: > I think `Type` was a much better suffix, the fact that it's a protocols is already encoded in the type.

To be fair, the fact that it’s a type is too 😁

> (Not critiquing this change though, I know it's aligned with the new stdlib conventions.)

Understood 👍 I was mostly curious cos I saw the opposite reaction from some folks when they moved to saying `FooType` in the first place.

norio-nomura:
Sorry, I didn't know that migration states are saved on each targets of Xcode project.
I updated `project.pbxproj`.

Result.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'Result'
-  s.version      = '2.1.1'
+  s.version      = '3.0.0-alpha.30p1'
```
mdiep: This should probably be `3.0.0-alpha.1`?

Result.podspec
```diff
@@ -9,7 +9,6 @@ Pod::Spec.new do |s|
   s.source       = { :git => 'https://github.com/antitypical/Result.git', :tag => s.version }
   s.source_files  = 'Result/*.swift'
   s.requires_arc = true
-  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '2.3' }
```
mdiep: This should probably explicitly specify 3.0 (or whatever the correct value is) for the `SWIFT_VERSION`.

Result/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.1.1</string>
+	<string>3.0.0-alpha.30p1</string>
```
mdiep: As above, I think `3.0.0-alpha.1` is probably more appropriate.

Result/Result.swift
```diff
@@ -262,6 +173,26 @@ extension NSError: ErrorTypeConvertible {
...
+		fatalError()
+	}
+}
+
+extension NSError {
+	@available(*, unavailable, renamed: "error(from:)")
+	public static func errorFromErrorType(_ error: ErrorProtocol) -> Self {
+		fatalError()
+	}
+}
```
mdiep: 💖

mdiep:
I think we should choose a simpler version number, as noted above.

And the tests didn't pass, but I think that may be because Travis has had problems today.

Once those are addressed, I'm 👍 to ship!

norio-nomura:
Thanks all for reviews!
I updated.

mdiep:
💖

norio-nomura:
Thanks! 🙏

PR #177, opened 2016-07-27
[Travis] Use `build-for-testing` and `test-without-building` build actions

ikesyo:
This should stabilize running iOS and tvOS tests on Travis with Xcode 8.


mdiep:
🚢 

Any idea why this works better?

ikesyo:
I'm not sure for the reasons. 😅

PR #182, opened 2016-08-17
Use Swift 3 precedence groups

mdiep:
Fixes the two warnings


Result/ResultProtocol.swift
```diff
@@ -120,14 +114,13 @@ public func &&& <L: ResultProtocol, R: ResultProtocol> (left: L, right: @autoclo
...
 
-infix operator >>- {
-	// Left-associativity so that chaining works like you’d expect, and for consistency with Haskell, Runes, swiftz, etc.
-	associativity left
-
-	// Higher precedence than function application, but lower than function composition.
-	precedence 100
+precedencegroup ChainingPrecedence {
+	associativity: left
+	higherThan: TernaryPrecedence
```
mdiep: I wasn't at all sure about the name or the higher/lower-than, but tried to pick what I thought made sense based on the docs.

NachoSoto:
:ship it:

jpsim:
👍. Thanks for dealing with the black art that is operator precedence. You could also remove the `UserInfoType` that used to handle the differences between Darwin/Linux bridging, since that's always `Any` now.

mdiep:
I think we need Swift 3.0 Preview 6 in order to make the tests pass. Gonna wait for that to drop before continuing with this PR. (Thankfully these are just warnings on `master`.)

norio-nomura:
> I think we need Swift 3.0 Preview 6

Swift 3.0 Preview 6 has come.
https://swift.org/download/#previews

Result/Result.swift
```diff
@@ -38,6 +38,7 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
 
 	// MARK: Deconstruction
 
+	#if !os(Linux)
```
jpsim: I filed [SR-2377](https://bugs.swift.org/browse/SR-2377) for this.

jpsim: Fixed in Swift master: https://github.com/apple/swift/pull/4353

PR #183, opened 2016-08-17
DEVELOPMENT-SNAPSHOT-2016-08-15-a

ikesyo:
This should pass the tests. 🙏


mdiep:
It looks like the Linux version doesn't work due to a swift error. 😞

PR #187, opened 2016-09-08
3.0.0-alpha.4 Podspec validation

juliengdt:
Since 3.0.0-alpha.4 has been released on origin/master, the wrong tag is on the podspec

Just applying good "s.version" for pod fetching


mdiep:
Oof. Thanks!

PR #193, opened 2016-10-26
Use Swift 3.0.1

jbrayton:
I updated the project configuration to use Swift 3.0.1.

### DIFF
--- a/Result.xcodeproj/project.pbxproj
+++ b/Result.xcodeproj/project.pbxproj
@@ -624,7 +624,7 @@
 				PRODUCT_NAME = Result;
 				SDKROOT = appletvos;
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 3.0;
+				SWIFT_VERSION = 3.0.1;
 				TARGETED_DEVICE_FAMILY = 3;
 			};
 			name = Debug;
@@ -647,7 +647,7 @@
 				PRODUCT_NAME = Result;
 				SDKROOT = appletvos;
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 3.0;
+				SWIFT_VERSION = 3.0.1;
 				TARGETED_DEVICE_FAMILY = 3;
 				VALIDATE_PRODUCT = YES;
 			};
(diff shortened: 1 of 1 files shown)


norio-nomura:
This changes make Result unbuildable by Xcode 8.0 with following error:
> “Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.

Are there any reason for requiring this changes?

Result.xcodeproj/project.pbxproj
```diff
@@ -816,6 +816,7 @@
 				PRODUCT_BUNDLE_IDENTIFIER = "com.antitypical.$(PRODUCT_NAME:rfc1034identifier)";
 				SDKROOT = macosx;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 3.0.1;
```
NachoSoto: This shouldn't be duplicated in the targets

ikesyo:
> This changes make Result unbuildable by Xcode 8.0 with following error:

Thanks @norio-nomura! Without this change, we can still build the framework with Xcode 8.1. So this should be closed I think.

ikesyo:
Closing this with the reasoning as I said above.

jbrayton:
I'm having trouble importing Result into a Swift 3.0.1 project, but since apparently no one else is seeing this (and because I'm not sure why I don't see it with other projects) I agree it should be closed for now. Sorry for the noise.

jbrayton:
The reason for the pull request is that I cannot use Result as a Carthage dependency with Xcode 8.1 (8T61a) and Swift 3.0.1.

Here is a sample project demonstrating the problem:

https://www.dropbox.com/s/gjzfvkvl2czeba0/ResultWithCarthage.zip?dl=0

It is an Xcode project that only varies from the "Single View Application" iOS template as follows:
* It incorporates Result as a Carthage dependency.
* The "AppDelegate.swift" file includes "import Result" at the top.

To reproduce the problem:

1.  Download the project.
2.  Run "carthage update" from the project directory.
3.  Try to run the project using Xcode 8.1 (8T61a) on a device or simulator.

You will see the following error:

Module compiled with Swift 3.0 cannot be imported in Swift 3.0.1.

To see the fix:

1.  Open Cartfile and comment out this:
github "antitypical/Result" "3.0.0"
2.  Uncomment this:
github "jbrayton/Result" "58a860424cda4742b485c412206c9efe62068c78"
3.  Clean and run the project.

That said, I understand that Xcode 8.0 is the version available in the Mac App Store and that this does not work in Xcode 8.0. I can appreciate why it may be best to hold off on this change.

Thanks.

jbrayton:
Oh! That works. Thank you, @ikesyo.

PR #198, opened 2016-11-14
Add type-erased `AnyError` for generic contexts

ikesyo:
In Swift 3, `Error` should be widely used instead of `NSError`. So the type should be useful to wrap arbitrary `Error` instances for uses in generic contexts.

This is an additive, non-breaking change.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -108,10 +108,26 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
 
 // MARK: - Derive result from failable closure
 
+public func materialize<T>(_ f: () throws -> T) -> Result<T, AnyError> {
+	return materialize(try f())
+}
+
+public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, AnyError> {
+	do {
+		return .success(try f())
+	} catch let error as AnyError {
+		return .failure(error)
+	} catch {
+		return .failure(AnyError(error))
+	}
+}
+
+@available(*, deprecated, message: "Use the overroad which returns `Result<T, AnyError>` instead")
 public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
 	return materialize(try f())
 }
 
+@available(*, deprecated, message: "Use the overroad which returns `Result<T, AnyError>` instead")
 public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, NSError> {
 	do {
 		return .success(try f())
@@ -160,7 +176,7 @@ extension NSError: ErrorProtocolConvertible {
 	}
 }
 
-// MARK: -
+// MARK: - Errors
 
 /// An “error” that is impossible to construct.
 ///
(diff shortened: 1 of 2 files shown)


ikesyo:
See also https://github.com/apple/swift-package-manager/blob/swift-DEVELOPMENT-SNAPSHOT-2016-11-12-a/Sources/Basic/Result.swift#L62-L74.

Result/Result.swift
```diff
@@ -108,10 +108,26 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
...
+		return .failure(AnyError(error))
+	}
+}
+
+@available(*, deprecated, message: "Use the overroad which returns `Result<T, AnyError>` instead")
 public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
 	return materialize(try f())
 }
 
+@available(*, deprecated, message: "Use the overroad which returns `Result<T, AnyError>` instead")
```
ikesyo: I'm not confident whether those should be deprecated or not.

hirohitokato: isn't it over'l'oad?

Result/Result.swift
```diff
@@ -169,6 +185,33 @@ extension NSError: ErrorProtocolConvertible {
...
+	public static func error(from error: Error) -> AnyError {
+		if let anyError = error as? AnyError {
+			return anyError
+		}
+		return AnyError(error)
+	}
+
+	public static func ==(lhs: AnyError, rhs: AnyError) -> Bool {
+		return lhs.error._code == rhs.error._code
+			&& lhs.error._domain == rhs.error._domain
```
norio-nomura: `_domain` should be compared before `_code`.

norio-nomura: Sorry, I withdraw my comment.

Result/Result.swift
ikesyo: 😱 Thank you!

Result/Result.swift
inamiy: Quick question: Are `_domain` and `_code` allowed to use?

ikesyo: > Are `_domain` and `_code` allowed to use?

I'm not confident. It should be okay to remove `Equatable` conformance (in that case use them for testing purpose only).

inamiy: > I'm not confident. It should be okay to remove Equatable conformance (in that case use them for testing purpose only).

+1. I also think removing `Equatable` is safer to ship :shipit:

ikesyo: Addressed in 3cafe3e.

ikesyo:
Any other thoughts? @antitypical/result

Result/Result.swift
```diff
@@ -169,6 +183,29 @@ extension NSError: ErrorProtocolConvertible {
 /// contains an `Int`eger and is guaranteed never to be a `failure`.
 public enum NoError: Swift.Error { }
 
+/// A type-erased error which wraps an arbitrary error instance. This should be
+/// useful for generic contexts.
+public struct AnyError: Swift.Error, ErrorProtocolConvertible, CustomStringConvertible {
```
NachoSoto: Maybe separate the conformances from the type?

Result/Result.swift
```diff
@@ -169,6 +183,29 @@ extension NSError: ErrorProtocolConvertible {
...
+	/// The underlying error.
+	public let error: Swift.Error
+
+	public init(_ error: Swift.Error) {
+		if let anyError = error as? AnyError {
```
NachoSoto: Nice

Result/Result.swift
ikesyo: Addressed in aac425d

ikesyo:
Thanks for the review comments! I'll merge this once CI passes, then cut a new release.

davidgreens:
@davidgreens

Result/Result.swift
davidgreens: Oh Yea.. @davidgreens

Result/Result.swift
davidgreens: Yep, 😎😷👏

PR #200, opened 2016-11-24
Add Equatable conformance to NoError

natestedman:
To allow use with functions dependent on `Equatable` conformance, i.e.

    let foo = Result<Int, NoError>(value: 10)
    let bar = Result<Int, NoError>(value: 10)

    foo == bar


mdiep:
This seems fine to me. Anyone have a reason why we shouldn't?

ikesyo:
Is this intentionally targeted to `swift2.x` branch, not `master` (Swift 3) ?

mdiep:
I'm embarrassed to say that I didn't notice which branch it was targeting. 😞

Issue #203, opened 2016-12-19
AnyError does not pass through localizedDescription from an underlying NSError


ratkins:
My intuition is that the below test should pass, but it does not in release 3.1.0:

```
class AnyErrorSpec: QuickSpec {
    override func spec() {
        describe("AnyError") {
            it("forwards an underlying NSError's localizedDescription") {
                let error = NSError(domain: "Foo", code: 42, userInfo: [NSLocalizedDescriptionKey: "a localized description"])

                expect(AnyError(error).localizedDescription).to(equal("a localized description"))
            }
        }
    }
}

```

Should it? Would you accept a PR to make it so?

ikesyo:
I expect you get the underlying error: `AnyError(error).error`. `localizedDescription` is not a requirement of `Error` protocol so `AnyError` does not implement it.

ratkins:
If `localizedDescription` isn't a property of `Error` (and you're right, it's not), why does this even parse?

```
let foo: AnyError = AnyError(error)
let bar: String = foo.localizedDescription

```

ikesyo:
See https://github.com/apple/swift-evolution/blob/master/proposals/0112-nserror-bridging.md

> In Foundation, add an extension to `ErrorProtocol` that provides access to the localized description, which is available for all error types.

ratkins:
Aha, I see, so implementing this for Result's `AnyError` would introduce a dependency on Foundation which you probably don't want to do. We can just add this locally then.

ikesyo:
It might make sense to conform to `LocalizedError` protocol and delegate to the underlying error.

PR #210, opened 2017-01-03
Add conformance to LocalizedError for AnyError

ratkins:
Fixes https://github.com/antitypical/Result/issues/203

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -214,6 +214,11 @@ extension AnyError: CustomStringConvertible {
 	}
 }
 
+// There appears to be a bug in Foundation on Linux which prevents this from working:
+// https://bugs.swift.org/browse/SR-3565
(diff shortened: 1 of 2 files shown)


ratkins:
Ok, that didn't go well. But I have no idea what the problem might be.

ratkins:
Now I have some idea what the problem is: [a bug](https://bugs.swift.org/browse/SR-3565) in either the linux version of Foundation or SwiftGlibc.

Would you accept this PR with the implementation `#ifdef`'d out on Linux until the above is fixed?

mdiep:
> Would you accept this PR with the implementation #ifdef'd out on Linux until the above is fixed?

I think that's acceptable. 👍 Definitely include a comment that links to the the bug.

mdiep:
@ikesyo Would you mind giving this another look? I think you're probably better qualified to judge it.

PR #213, opened 2017-02-07
Change to use `@available(swift, deprecated: 3.0, obsoleted: 3.1, …`

### LINKED ISSUE #212: Build failed on Linux with swift 3.1
Sorry to bother you guys with the pre-released swift version. I just want you to know this issue.

The issue was reported [here](https://bugs.swift.org/browse/SR-3872). And the apple guys confirmed the issue is intended behaviour. The related code is [here](https://github.com/antitypical/Result/blob/master/Result/Result.swift#L128-L135).

This actually won't be an issue once we remove these deprecated methods.

norio-nomura:
Fixes #212

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,10 @@ matrix:
       language: generic
       install:
         - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
+    - script:
+        - docker run -v `pwd`:`pwd` -w `pwd` norionomura/swift:3120170205a swift test
+      env: JOB=Linux-Swift3.1
+      sudo: required
+      services: docker
 notifications:
   email: false
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -122,16 +122,12 @@ public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, AnyErr
 
 @available(*, deprecated, message: "Use the overload which returns `Result<T, AnyError>` instead")
 public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
-	return materialize(try f())
+	fatalError("unreachable")
 }
 
 @available(*, deprecated, message: "Use the overload which returns `Result<T, AnyError>` instead")
 public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, NSError> {
-	do {
-		return .success(try f())
-	} catch let error as NSError {
-		return .failure(error)
-	}
+	fatalError("unreachable")
 }
 
 // MARK: - Cocoa API conveniences
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -120,35 +120,28 @@ public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, AnyErr
 	}
 }
 
+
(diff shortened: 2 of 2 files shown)


norio-nomura:
Should we keep them workable?

ikesyo:
Since this is breaking API change now, I think we should choice either of the following:

- Make `master` branch for Swift 3.1
- Prepare `master-next` for Swift 3.1 and `master` is still for Swift 3.0 until we get the official Xcode 8.3 /  Swift 3.1 release.

I prefer the latter.

norio-nomura:
How do you think this?:
```swift
@available(*, deprecated, message: "Use the overload which returns `Result<T, AnyError>` instead")
public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, NSError> {
	#if os(Linux)
		#if swift(>=3.1)
			fatalError("unreachable")
		#else
			do {
				return .success(try f())
			} catch let error as NSError {
				return .failure(error)
			}
		#endif
	#else
		do {
			return .success(try f())
		} catch let error as NSError {
			return .failure(error)
		}
	#endif
}
```

norio-nomura:
Or 
```swift
#if swift(>=3.1)

@available(*, unavailable, message: "Use the overload which returns `Result<T, AnyError>` instead")
public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
	fatalError("unreachable")
}

@available(*, unavailable, message: "Use the overload which returns `Result<T, AnyError>` instead")
public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, NSError> {
	fatalError("unreachable")
}

#else

@available(*, deprecated, message: "Use the overload which returns `Result<T, AnyError>` instead")
public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
	return materialize(try f())
}

@available(*, deprecated, message: "Use the overload which returns `Result<T, AnyError>` instead")
public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, NSError> {
	do {
		return .success(try f())
	} catch let error as NSError {
		return .failure(error)
	}
}

#endif
```

norio-nomura:
Maybe we should add job to CI with next snapshot of Swift 3.1.

norio-nomura:
Job added using 'swift-3.1-DEVELOPMENT-SNAPSHOT-2017-02-05-a` to Travis

ikesyo:
Even if we conditionalize the unavailability only for Swift 3.1 on Linux, that is still API breaking (on the platform) and requires a major version bump (Result 4.x). So let's simply mark them unavailable for all platforms unconditionally and merge it into `master-next` branch, which is dedicated for Swift 3.1 changes. And leave the `master` branch as is (for Swift 3.0.x) for a while until we get Xcode 8.3 GM.

Does it make sense?

norio-nomura:
I have removed `#if os(Linux)` from them.
I updated to use `@available(swift, deprecated: 3.0, obsoleted: 3.1, …` that would be intended to use on advance preparations of deprecation.
Same usage can be found in https://github.com/apple/swift/blob/swift-3.1-DEVELOPMENT-SNAPSHOT-2017-02-05-a/stdlib/public/SDK/CoreGraphics/CGFloat.swift.gyb#L339

norio-nomura:
Event if we would use `master-next` for Result-4.x, I guess this PR could fit to that. 🙂

norio-nomura:
I changed base branch to `master-next` and title.

norio-nomura:
Thanks! 🙏

PR #214, opened 2017-03-01
[PoC] Modify `materialize` overload of `Result<T, NSError>` to be buildable on both Swift 3.0 and 3.1

ikesyo:
This is an alternative approach to address #212 (see also #213). This does not require major version bump so make it easy to upgrade to Swift 3.1. But I'm not confident if this is acceptable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,10 @@ matrix:
       language: generic
       install:
         - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
+    - script:
+        - docker run -v `pwd`:`pwd` -w `pwd` norionomura/swift:3120170205a swift test
+      env: JOB=Linux-Swift3.1
+      sudo: required
+      services: docker
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -29,5 +29,10 @@ matrix:
       language: generic
       install:
         - eval "$(curl -sL https://gist.githubusercontent.com/kylef/5c0475ff02b7c7671d2a/raw/9f442512a46d7a2af7b850d65a7e9bd31edfb09b/swiftenv-install.sh)"
+    - script:
+        - docker run -v `pwd`:`pwd` -w `pwd` norionomura/swift:3120170205a swift test
```
mdiep: This deserves a comment as I'm not really sure what it's doing.

Result/Result.swift
```diff
@@ -129,8 +129,15 @@ public func materialize<T>(_ f: () throws -> T) -> Result<T, NSError> {
...
-		return .failure(error)
+	} catch {
+#if _runtime(_ObjC)
+		return .failure(error as NSError)
+#else
+		// https://github.com/apple/swift-corelibs-foundation/blob/swift-3.0.2-RELEASE/Foundation/NSError.swift#L314
+		let userInfo = _swift_Foundation_getErrorDefaultUserInfo(error) as? [String: Any]
+		let nsError = NSError(domain: error._domain, code: error._code, userInfo: userInfo)
+		return .failure(nsError)
+#endif
```
mdiep: I think this is acceptable, but would prefer to see a comment that says something like "This isn't great, but it lets us maintain compatibility until this deprecated method can be removed."

.travis.yml
ikesyo: Replaced it with a recent 3.1-DEVELOPMENT-SNAPSHOT on Linux: b26e951

PR #217, opened 2017-03-30
Let's move to Swift 3.1 world

ikesyo:
Once this is merged, I'd like to remove some deprecated APIs.


Result.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'Result'
-  s.version      = '3.2.1'
+  s.version      = '4.0.0-alpha.1'
```
norio-nomura: Swift Package Manager in Swift 3.1 would not work with pre-release tag.
https://bugs.swift.org/browse/SR-1039

ikesyo: This is just a placeholder and not for SwiftPM.

PR #220, opened 2017-04-25
Remove deprecated APIs

### LINKED ISSUE #191: Avoid using custom operators in Swift 3
It seems `>>-` operator no longer works seamlessly with other operator-centric libraries e.g. [Runes](https://github.com/thoughtbot/Runes), [Operadics](https://github.com/typelift/Operadics), due to the new `precedencegroup` in Swift 3. 

Formerly in Swift 2, `"error: ambiguous operator declarations found for operator"` did not occur iff `associativity` and `precedence` were totally equal, but this workaround no longer works in Swift 3, so I would like to suggest to either:
1. Remove custom operators in `Result`, or
2. Add de facto infix-operator library as a dependency (if there's any)

ikesyo:
Closes #191.

### DIFF
--- a/Result/ResultProtocol.swift
+++ b/Result/ResultProtocol.swift
@@ -121,31 +121,6 @@ public extension ResultProtocol where Error: ErrorProtocolConvertible {
 
 // MARK: - Operators
 
-infix operator &&& : LogicalConjunctionPrecedence
-
-/// Returns a Result with a tuple of `left` and `right` values if both are `Success`es, or re-wrapping the error of the earlier `Failure`.
-@available(*, deprecated, renamed: "ResultProtocol.fanout(self:_:)")
-public func &&& <L: ResultProtocol, R: ResultProtocol> (left: L, right: @autoclosure () -> R) -> Result<(L.Value, R.Value), L.Error>
-	where L.Error == R.Error
-{
-	return left.fanout(right)
-}
-
-precedencegroup ChainingPrecedence {
-	associativity: left
-	higherThan: TernaryPrecedence
-}
-
-infix operator >>- : ChainingPrecedence
-
-/// Returns the result of applying `transform` to `Success`es’ values, or re-wrapping `Failure`’s errors.
-///
-/// This is a synonym for `flatMap`.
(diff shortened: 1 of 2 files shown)


Result/ResultProtocol.swift
```diff
@@ -121,31 +121,6 @@ public extension ResultProtocol where Error: ErrorProtocolConvertible {
...
-	associativity: left
-	higherThan: TernaryPrecedence
-}
-
-infix operator >>- : ChainingPrecedence
-
-/// Returns the result of applying `transform` to `Success`es’ values, or re-wrapping `Failure`’s errors.
-///
-/// This is a synonym for `flatMap`.
-@available(*, deprecated, renamed: "ResultProtocol.flatMap(self:_:)")
```
mdiep: Can we mark these as `unavailable` as well?

mdiep:
I suppose we should look into removing `ResultProtocol` and doing a 4.0 release?

Thomvis:
@mdiep I proposed adding the `ResultProtocol` (a long time ago) to be able to do https://github.com/Thomvis/BrightFutures/blob/master/Sources/BrightFutures/AsyncType%2BResultType.swift#L10. But that no longer requires a protocol in Swift 3.1, right?

Result/ResultProtocol.swift
ikesyo: The operator definitions must be remained to do that. However, we actually want to remove the operators to close #191.

mdiep:
> But that no longer requires a protocol in Swift 3.1, right?

Correct

ikesyo:
@Thomvis @mdiep I think for that purpose [Parameterized extensions](https://github.com/apple/swift/blob/master/docs/GenericsManifesto.md#parameterized-extensions) is needed since `Result` itself is generic, which is not implemented yet.

ikesyo:
@mdiep @Thomvis Do you have further concerns? I'd like to have Result 2.0 with these changes for the forthcoming ReactiveSwift 2.0.

andersio:
The protocol can be reduced to return the concrete type like `OptionalProtocol` in RAS, and have all methods moved to the concrete type.

PR #223, opened 2017-05-15
Improve AnyError support in `Result.init(attempt:)`

ikesyo:
Originally reported in https://github.com/mdiep/Tentacle/issues/66.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -31,7 +31,10 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
 	public init(attempt f: () throws -> T) {
 		do {
 			self = .success(try f())
-		} catch {
+		} catch var error {
+			if Error.self == AnyError.self {
+				error = AnyError(error)
+			}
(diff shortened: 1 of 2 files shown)


Result/Result.swift
```diff
@@ -31,7 +31,10 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
 	public init(attempt f: () throws -> T) {
 		do {
 			self = .success(try f())
-		} catch {
+		} catch var error {
+			if Error.self == AnyError.self {
```
mdiep: This should probably only be if `error` _isn't_ an `AnyError`? (Can you add a test for that?)

ikesyo: I think that is already covered here: https://github.com/antitypical/Result/blob/65d0c81e295e3781c9fea5bec309be049b045a3a/Result/Result.swift#L183-L187

mdiep: 👍 

I didn't realize that.

ikesyo:
Thanks for the review! 🚢

PR #232, opened 2017-05-29
Rename ErrorProtocolConvertible to ErrorConvertible

ikesyo:
Ports #230 into `master` and makes the old name unavailable.

### DIFF
--- a/Result/ResultProtocol.swift
+++ b/Result/ResultProtocol.swift
@@ -98,11 +98,11 @@ public extension ResultProtocol {
 }
 
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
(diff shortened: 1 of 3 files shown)


Result/ResultProtocol.swift
```diff
@@ -98,11 +98,11 @@ public extension ResultProtocol {
 }
 
 /// Protocol used to constrain `tryMap` to `Result`s with compatible `Error`s.
-public protocol ErrorProtocolConvertible: Swift.Error {
+public protocol ErrorConvertible: Swift.Error {
 	static func error(from error: Swift.Error) -> Self
```
andersio: Could this be an initialiser?

ikesyo: Let's explore it in a separete pull request.

PR #233, opened 2017-05-29
Convert `materialize` to be a `Result` initializer.

andersio:


Tests/ResultTests/ResultTests.swift
```diff
@@ -120,21 +120,26 @@ final class ResultTests: XCTestCase {
...
 		XCTAssert(result1.error == error)
 
-		let result2: Result<String, AnyError> = materialize { try tryIsSuccess(nil) }
+		let result2: Result<String, AnyError> = Result { try tryIsSuccess(nil) }
 		XCTAssert(result2.error == error)
 	}
 
+	func testMaterializeInferrence() {
+		let result = Result { try tryIsSuccess(nil) }
+		XCTAssert((type(of: result) as Any.Type) is Result<() throws -> String, AnyError>.Type)
```
andersio: Do note that the inferred `T` type is the same as `materialize`, just that it would be ambiguous due to having both the `AnyError` and `NSError` overloads.

Result/Result.swift
```diff
@@ -109,18 +116,25 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
...
+	public init(_ f: () throws -> T) {
+		self.init(_attempt: f)
+	}
 
+	public init(_ f: @autoclosure () throws -> T) {
+		self.init(_attempt: f)
+	}
+}
+
+@available(*, deprecated, renamed: "Result.init(attempt:)")
```
mdiep: That looks like the wrong renamed method.

Result/Result.swift
```diff
@@ -24,11 +24,18 @@ public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertib
 
 	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
 	public init(_ f: @autoclosure () throws -> T) {
-		self.init(attempt: f)
+		self.init(_attempt: f)
 	}
 
-	/// Constructs a result from a function that uses `throw`, failing with `Error` if throws.
+	@available(*, deprecated, renamed:"init(_:)")
 	public init(attempt f: () throws -> T) {
```
mdiep: This doesn't seem equivalent to `init(_:)` because the latter is in a constrained extension where `Error == AnyError`.

So before you could write `Result<Int, NSError>(attempt: f)`, but now there's no equivalent.

mdiep:
Closing in favor of #236.

PR #234, opened 2017-05-29
Remove `ResultProtocol`.

andersio:
No operator requires it, and AFAIU `Result` is built on the assumption/premises of providing a currency type rather than defining a generic protocol.


andersio:
Just read #220.

PR #235, opened 2017-05-29
Reduce the responsibility of `ResultProtocol`.

andersio:
Similar to signal and producer operators in ReactiveSwift, method and computed properties are now moved to the concrete type. `ResultProtocol` now requires only a `ResultProtocol.result` requirement which materializes `self`.

Caveat: Automatic migration is impossible, but manual migration is not hard (copy & pasting `.result`).

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -1,14 +1,14 @@
 //  Copyright (c) 2015 Rob Rix. All rights reserved.
 
 /// An enum representing either a failure with an explanatory error, or a success with a result value.
-public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertible, CustomDebugStringConvertible {
-	case success(T)
+public enum Result<Value, Error: Swift.Error>: ResultProtocol, CustomStringConvertible, CustomDebugStringConvertible {
+	case success(Value)
 	case failure(Error)
 
 	// MARK: Constructors
 
 	/// Constructs a success wrapping a `value`.
-	public init(value: T) {
+	public init(value: Value) {
 		self = .success(value)
 	}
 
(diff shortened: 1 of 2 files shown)


Result/Result.swift
```diff
@@ -1,14 +1,14 @@
 //  Copyright (c) 2015 Rob Rix. All rights reserved.
 
 /// An enum representing either a failure with an explanatory error, or a success with a result value.
-public enum Result<T, Error: Swift.Error>: ResultProtocol, CustomStringConvertible, CustomDebugStringConvertible {
-	case success(T)
+public enum Result<Value, Error: Swift.Error>: ResultProtocol, CustomStringConvertible, CustomDebugStringConvertible {
```
mdiep: This is a separate, breaking change. We should probably do this separately?

Result/ResultProtocol.swift
```diff
@@ -1,34 +1,14 @@
 //  Copyright (c) 2015 Rob Rix. All rights reserved.
 
-/// A type that can represent either failure with an error or success with a result value.
+// A protocol that can be used to constrain associated types as `Result`.
```
mdiep: This should have 3 `/`s so it shows up in Xcode's documentation viewer.

mdiep:
I'd definitely love to have @ikesyo review this. He has a much better understanding of the consequences of this than I do.

Result/Result.swift
mdiep: Since we're all for this change, so I'm okay with just doing it here.

Result/ResultProtocol.swift
ikesyo: 5baafe6

mdiep:
🤘

PR #236, opened 2017-06-12
Add AnyError overloads for init(_:) and init(attempt:)

ikesyo:
Similar to but slightly different from #233 (and smaller diff). I prefer to have separate signatures for each `init` overloads.

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -114,18 +114,30 @@ public enum Result<Value, Error: Swift.Error>: ResultProtocol, CustomStringConve
 	}
 }
 
+extension Result where Error == AnyError {
+	/// Constructs a result from an expression that uses `throw`, failing with `AnyError` if throws.
+	public init(_ f: @autoclosure () throws -> Value) {
+		self.init(attempt: f)
+	}
+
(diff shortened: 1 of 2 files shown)


Result/Result.swift
```diff
@@ -114,18 +114,30 @@ public enum Result<Value, Error: Swift.Error>: ResultProtocol, CustomStringConve
...
+
+	/// Constructs a result from a closure that uses `throw`, failing with `AnyError` if throws.
+	public init(attempt f: () throws -> Value) {
+		do {
+			self = .success(try f())
+		} catch {
+			self = .failure(AnyError(error))
+		}
+	}
+}
```
mdiep: Are these overloads necessary?

The changes below compile without this change. I'm not sure what this adds.

(Type inference PRs are some of the hardest to review! 😅)

ikesyo: I've just added https://github.com/antitypical/Result/pull/236/files#diff-edf3d2a3570d3672e50cbe6ca3ede069R140 and without these overloads that does not compile.

The test case / result are bit different from #233.

PR #242, opened 2017-07-18
Move to swift 3.2

AltiAntonov:


AltiAntonov:
We need this, otherwise we cannot use the library on xCode 9 and Swift 4.(xCode 9 read only swift 3.2 and 4)

PR #243, opened 2017-07-23
Add initializers to ResultProtocol

Thomvis:
Hi! I'd like to propose adding back the initializers to `ResultProtocol`. This will allow creation of `ResultProtocol` 'instances'  in contexts where it is used as a type constraint. More specifically, I would need it to be able to upgrade BrightFutures to use Result `master`. For example usage see, https://github.com/Thomvis/BrightFutures/blob/master/Sources/BrightFutures/MutableAsyncType+ResultType.swift#L17.

I understand these were removed in #235, as part of an effort to reduce `ResultProtocol`'s responsibilities. I don't think I understand the reasoning behind that PR, so I apologise if I'm missing something obvious.

Thanks!

### DIFF
--- a/Result/ResultProtocol.swift
+++ b/Result/ResultProtocol.swift
@@ -5,6 +5,9 @@ public protocol ResultProtocol {
 	associatedtype Value
 	associatedtype Error: Swift.Error
 
+	init(value: Value)
+	init(error: Error)
+	
 	var result: Result<Value, Error> { get }
 }
 


mdiep:
`ResultProtocol` has always been a workaround for Swift's incomplete generics system. It's not intended to be a protocol that types conform to—it's only meant to allow you to constrain items where Swift doesn't currently allow constraining to `Result` itself. Once these limitations are removed, we'll remove `ResultProtocol` entirely.

Having said that, I think we do need to add these back so that you constrain to a generic `Result` today. Thanks for the pull request!

PR #248, opened 2017-09-08
Merge multiple targets into single universal target

regexident:
With Apple collecting ever more device types and platforms in its portfolio it is getting increasingly more tedious and harder to scale the development of multi-platform frameworks by using dedicated platform targets.

The use of an `.xcconfig` file allows for development of frameworks for an arbitrary number of platforms with just a single universal framework target.

Universal frameworks have the additional benefit of being linkable from single-platform targets as well as universal targets:

Compatibility|Platform Host|Universal Host
-----|-----|-----
**Platform Framework**|✅|💥
**Universal Framework**|✅|✅

I would therefor like to propose changing **Result** to use such an `.xcconfig`, making it usable in projects with single-platform targets as well as universal ones.


mdiep:
I don't personally feel a need for this, but I'm curious what the rest of @antitypical/result thinks.

regexident:
If you're happy with **Result** being somewhat of a pain to link against in projects with universal targets, then certainly so. Otherwise however, … :wink:

ikesyo:
> but I'm curious what the rest of @antitypical/result thinks.

I'm mot a fan of this either.

regexident:
> I'm mot a fan of this either.

Curious. Mind to elaborate why (other than personal taste)?

Issue #85, opened 2015-09-19
Remove `try`


neilpa:
Now that Cocoa APIs map to `throws` we should be able to remove these and rely entirely on `materialize`.

neilpa:
Although, it may be interesting to explore repurposing these for C-APIs returning error codes and out-params.

robrix:
Those can be done orthogonally. :+1: to both.

neilpa:
Opened #87 for the C wrappers

PR #251, opened 2017-09-26
[3.x] Deprecate `try` functions

ikesyo:
Those will be removed in Result 4.0.

This is a preparation for #85.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
         - swift test
       env:
         - JOB=Linux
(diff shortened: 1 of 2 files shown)


gfontenot:
Do we maybe want to point people towards `materialize` in the message?

ikesyo:
Thanks @gfontenot and @ashfurrow. Since the `materialize` itself is marked as deprecated on `master`, I pointed to `Result.init(attempt:)` instead.

https://github.com/antitypical/Result/blob/39e761c/Result/Result.swift#L136-L139

gfontenot:
Ah, right. 👍

PR #252, opened 2017-09-27
[4.0] Drop Xcode 8.3/Swift 3.1 support

ikesyo:
We now require Xcode 9/Swift 4 compiler.


ikesyo:
I'll go ahead. :shipit:

PR #254, opened 2017-11-09
Throw other errors within `init(attempt:)` and `tryMap`

philtre:
When using Result in my project, I needed to use `init(attempt:)` in several places but I wanted to use my own wrapper for arbitrary `Error`s. The existing initializer only works with `AnyError`, so I've modified the code to allow using any `Error` which conforms to the `ErrorInitializing` protocol.


philtre:
Noticed that some tests were failing. Those errors are now fixed and I added a few more tests.

PR #257, opened 2017-12-10
`Swift.Error::anyError` computed property.

Vespen:
I think it is better than using `AnyError::init`, because it is not obvious (for API user) that `AnyError::init` will not produce double boxing for `AnyError` errors. And result syntax looks not bad `Future<..., AnyError>(error: error.anyError)`.

### DIFF
--- a/Result/AnyError.swift
+++ b/Result/AnyError.swift
@@ -44,3 +44,13 @@ extension AnyError: LocalizedError {
 		return (error as? LocalizedError)?.recoverySuggestion
 	}
 }
+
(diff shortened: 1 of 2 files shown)


mdiep:
Thanks for the PR!

We chosen to use an `init` for this because it's consistent with the Swift API guidelines and the Swift standard library, which use `init`s for type conversions. I don't think adding a second, non-standard way to do this would be a good idea.

ikesyo:
Closing this due to inactivity. @mdiep clearly explained our direction.

PR #258, opened 2018-01-17
Use Swift 4 and Xcode 9 in all targets

bricklife:
Fix Swift 4 conversion warning in each target.


bricklife:
@NachoSoto Thank you very much for your approval! Could you re-start the CI job at your convenience? It seems that Travis CI was down when I pushed this PR... 😢

norio-nomura:
restarted the CI job.

PR #264, opened 2018-04-26
Update podspec

ikesyo:
- Add `swift_version` attribute introduced in CocoaPods 1.4.0
- Add `cocoapods_version` attribute to require CocoaPods 1.4.0


mdiep:
Don't forget to update the `Result` version before the release. 😄

PR #266, opened 2018-06-07
Move Headers phases before Sources phases

mdiep:
I don't _think_ it's going to be a problem for this framework, but this is the correct way to order these. The Swift files depend on the framework header, so it should be operated on first.


NachoSoto:
👍

PR #267, opened 2018-06-22
Update README.md

valeriyvan:
Refactors first example using guard let instead of if let


NachoSoto:
I wouldn't say this is objectively better, I think the symmetry of the original code makes more sense.

mdiep:
I think this makes sense since there are multiple error conditions.

PR #270, opened 2018-09-16
Fix swift_version on podspec

giginet:
Just run `pod lib lint` on this repository, CocoaPods raises following error.

```
Result (master) ~ pod lib lint

 -> Result
 -> Result (4.0.0)
    - ERROR | swift: Specification `Result` specifies an inconsistent `swift_version` (`4.0`) compared to the one present in your `.swift-version` file (`4.1`). Please remove the `.swift-version` file which is now deprecated and only use the `swift_version` attribute within your podspec.

[!] Result did not pass validation, due to 1 error.
You can use the `--no-clean` option to inspect any issue.

```

Why did it mismatch between `podspec` and `.swift_version`?

I think it is strange.
Result 4.0 supports only above Swift 4.1, so `*.podspec` should be fixed.

### DIFF
--- a/Result.podspec
+++ b/Result.podspec
@@ -15,5 +15,5 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '9.0'
   
   s.swift_version = '4.1'
-  s.cocoapods_version = '>= 1.4.0'
+  s.cocoapods_version = '>= 1.5.0'
 end
--- a/Result.podspec
+++ b/Result.podspec
@@ -14,6 +14,6 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
   
-  s.swift_version = '4.0'
+  s.swift_version = '4.1'
   s.cocoapods_version = '>= 1.4.0'
 end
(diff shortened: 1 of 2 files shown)


Result.podspec
```diff
@@ -14,6 +14,6 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target = '9.0'
   
-  s.swift_version = '4.0'
-  s.cocoapods_version = '>= 1.4.0'
+  s.swift_version = '4.1'
```
ikesyo: This is for `SWIFT_LANGUAGE_VERSION` build setting, so 4.0 in Xcode 9.4 means Swift 4.1; it's not strange. But Xcode also accepts 4.1 as a valid value so this is okay.

Result.podspec
```diff
@@ -14,6 +14,6 @@ Pod::Spec.new do |s|
...
+  s.cocoapods_version = '>= 1.5.0'
```
ikesyo: Why is CocoaPods 1.5.0 required? 1.4.0 requirement is to use `s.swift_version` syntax. Is there addional requirment?

ikesyo:
I'd say that the lint error will not happen on CocoaPods 1.6.0.

ikesyo:
This should be addressed by #271. Closing.

Thanks for the PR though! 😄

PR #278, opened 2018-12-31
Add compatibility with Swift 5's Result

ikesyo:
This aims at providing a migration path for the Result type in the Swift 5's standard library.

- https://github.com/apple/swift-evolution/blob/master/proposals/0235-add-result.md
- https://forums.swift.org/t/accepted-with-modifications-se-0235-add-result-to-the-standard-library/18603

### DIFF
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -173,7 +173,7 @@ public func materialize<T>(_ f: () throws -> T) -> Result<T, AnyError> {
 
 @available(*, deprecated, renamed: "Result.init(_:)")
 public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, AnyError> {
-	return Result(f)
+	return Result(try f())
 }
 
 // MARK: - ErrorConvertible conformance
--- a/Result/Result.swift
+++ b/Result/Result.swift
@@ -5,6 +5,19 @@ public enum Result<Value, Error: Swift.Error>: ResultProtocol, CustomStringConve
 	case success(Value)
 	case failure(Error)
 
+	/// The compatibility alias for the Swift 5's `Result` in the standard library.
+	///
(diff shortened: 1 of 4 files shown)


norio-nomura:
Since `Package.swift` still uses `swift-tools-version:4.0`, SwiftPM uses `-swift-version 4` on compiling. It would be better to add `Package@swift-5.swift` using `swift-tools-version:5.0` for checking compatibility with `-swift-version 5`.

ikesyo:
@norio-nomura Thanks for pointing it out! I'll address it.

Result/Result.swift
```diff
@@ -140,7 +173,7 @@ public func materialize<T>(_ f: () throws -> T) -> Result<T, AnyError> {
 
 @available(*, deprecated, renamed: "Result.init(_:)")
 public func materialize<T>(_ f: @autoclosure () throws -> T) -> Result<T, AnyError> {
-	return Result(f)
+	return Result(try f())
```
ikesyo: This is needed (for Swift 5) because of the error: `error: add () to forward @autoclosure parameter`.

ikesyo:
Thanks!

mdiep:
Awesome. 🤘🏻

I'm assuming we'll do a 5.0 release that removes everything that's defined in the Standard Library version.

marchy:
@mdiep Do you anticipate this as a future release beyond the current type-aliased 4.1.0 release? Great to see the migration path here and the enablement which this library has produced  =)

mdiep:
I suppose we should do this, yes. But I'm unclear if this should be a minor release or a major release.

marchy:
Makes perfect sense. I suppose should go as a major release, even if the consuming APIs may not be breaking (which I'm not sure is the case or not).

Conceptually it's definitely a big thing, so a major release will at least capture/communicate that.

PR #279, opened 2019-01-02
Split implementation into some extensions

ikesyo:
This would make it easier to just extend stdlib Result on Swift 5.


mdiep:
Nice. 👏

PR #280, opened 2019-01-10
Use `Swift.Result` instead of our `Result` if `compiler(>=5)`

norio-nomura:


norio-nomura:
This makes it easier for users using antitypical/Result to switch to `Swift.Result`.

I confirmed that [Commandant](https://github.com/Carthage/Commandant) can be built with this and using `Swift.Result`:
`final func run(command verb: Swift.String, arguments: [Swift.String]) -> Swift.Result<(), Commandant.CommandantError<ClientError>>?`
as following:
```terminal.sh-session
$ docker run --privileged -it -v `pwd`:`pwd` -w `pwd` --rm norionomura/swift:swift-5.0-branch swift run --repl -Xswiftc -swift-version -Xswiftc 5
[1/1] Linking ./.build/x86_64-unknown-linux/debug/libCommandant__REPL.so
Launching Swift REPL with arguments: -I/Users/norio/github/swift-dev/SourceKitten/Carthage/Checkouts/Commandant/.build/x86_64-unknown-linux/debug -L/Users/norio/github/swift-dev/SourceKitten/Carthage/Checkouts/Commandant/.build/x86_64-unknown-linux/debug -lCommandant__REPL
Welcome to Swift version 5.0-dev (LLVM eb266bb301, Clang 2b49a6d7dc, Swift 5bf394d8ff).
Type :help for assistance.
  1> import Commandant
  2> :type lookup CommandRegistry
final class CommandRegistry<ClientError> where ClientError : Error {
  @_hasInitialValue final var commandsByVerb: [Swift.String : Commandant.CommandWrapper<ClientError>]
  final var commands: [Commandant.CommandWrapper<ClientError>] {
    get
  }
  init()
  @discardableResult final func register<C>(_ commands: C...) -> Commandant.CommandRegistry<ClientError> where ClientError == C.ClientError, C : CommandProtocol
  final func run(command verb: Swift.String, arguments: [Swift.String]) -> Swift.Result<(), Commandant.CommandantError<ClientError>>?
  final subscript(verb: Swift.String) -> Commandant.CommandWrapper<ClientError>? {
    get
  }
  deinit
}
extension CommandRegistry {
  final func main(defaultVerb: Swift.String, errorHandler: (ClientError) -> ()) -> Swift.Never
  final func main(arguments: [Swift.String], defaultVerb: Swift.String, errorHandler: (ClientError) -> ()) -> Swift.Never
  final func executeSubcommandIfExists(_ executableName: Swift.String, verb: Swift.String, arguments: [Swift.String]) -> Swift.Int32?
}

  2>  
```

Result/Result.swift
```diff
@@ -1,10 +1,35 @@
 //  Copyright (c) 2015 Rob Rix. All rights reserved.
 
+#if swift(>=4.2)
+#if compiler(>=5)
```
mdiep: I don't think this maintains the requirements for semantic version compatibility, so I think we need to drop support for Swift 4 to do this. (And we should do this!)

If we conditionally define the enum, then the output is different if Result is compiled in Swift 4 mode or Swift 5 mode. Since semantic versioning was defined by TPW and he writes Ruby, it's not totally clear whether it's meant to be source- or binary-compatible, but I think the safe thing is to require both.

mdiep:
I'm very excited to see this! 👏🏻

Result/Result.swift
norio-nomura: >  I think we need to drop support for Swift 4 to do this. (And we should do this!)

Just changing the major version is sufficient, and it seems there is no need to make it impossible to build with older Swift compiler.🤔 

> If we conditionally define the enum, then the output is different if Result is compiled in Swift 4 mode or Swift 5 mode.

It is not affected by the `-swift-version` flag. When using the Swift 5+ compiler, `Swift.Result` will be used.

norio-nomura: Sorry, edited previous comment. 🙇

mdiep: Oh, I see! Tricky!

norio-nomura:
Thanks! 🙏

PR #285, opened 2019-03-29
Fix Xcode 10.2 warnings

mdiep:


DivineDominion:
Is this ready to be merged and released in Result v5.0? Do you need help with anything?

ikesyo:
> Is this ready to be merged and released in Result v5.0?

This should not need major version bump.

DivineDominion:
Not on its own, but a v5 release that signals the transition to Swift 5's native Result type support would be nice! :)

PR #287, opened 2019-04-23
Deprecate NoError enum

nekrich:
Hi.

I suggest deprecating `NoError` in favor of native `Never` since it conforms to Error & Equatable.

My view of releases for smooth migration in this case:
v5.0 - Deprecate enum
v6.0 - Replace enum with deprecated type alias
v6.1 - Remove type alias

### DIFF
--- a/Result/NoError.swift
+++ b/Result/NoError.swift
@@ -3,8 +3,17 @@
 /// This can be used to describe `Result`s where failures will never
 /// be generated. For example, `Result<Int, NoError>` describes a result that
 /// contains an `Int`eger and is guaranteed never to be a `failure`.
+#if swift(>=5.0)
+@available(*, deprecated, message: "Use `Swift.Never` instead", renamed: "Never")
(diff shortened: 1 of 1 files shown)


Tests/ResultTests/NeverTests.swift
```diff
@@ -0,0 +1,13 @@
...
+
+#if swift(>=5.0)
+final class NeverTests: XCTestCase {
+	func testEquatable() {
+		let foo = Result<Int, Never>(1)
+		let bar = Result<Int, Never>(1)
+		XCTAssertTrue(foo == bar)
+	}
+}
+#endif
```
mdiep: This test is unnecessary. If `swift(>=5.0)`, then all of these types are from the standard library. We don't need to test the contents of the standard library.

Could you remove this test?

nekrich: Done

mdiep:
This is great! Thanks!

